PR #7, opened 2023-02-10
Account for LVGL task handler delay

GeorgeCGV:
The LVGL task handler returns the amount of time before the next update call.
This PR corrects the 'lvgl_render' work schedule to take that time into account.

That shall improve LVGL responsiveness.

### DIFF
--- a/app/src/main.c
+++ b/app/src/main.c
@@ -35,8 +35,6 @@
 
 LOG_MODULE_REGISTER(main, LOG_LEVEL_WRN);
 
-#define RENDER_INTERVAL_LVGL    K_MSEC(100)
-
 typedef enum ui_state {
     INIT_STATE,
     SETTINGS_STATE,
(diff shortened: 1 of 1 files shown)


jakkra:
Thanks, will try it out, however there is a reason I did it as I did. Because sometimes the call to `lv_task_handler` takes a LONG time, like 50-100 ms, if I then call it pretty much immediately again there is no time for other stuff in the system and no time to sleep. This is why I kept a fixed interval between the calls.

If `lv_task_handler` returns time to next, I guess there is somewhere a config of "target" update rate in LVGL I might have to modify?

GeorgeCGV:
I see. I assume that `LV_DISP_DEF_REFR_PERIOD` and `LV_INDEV_DEF_READ_PERIOD` shall be tweaked to reach desired performance. There is also `LV_LOG_TRACE_DISP_REFR` that can assist with figuring out what causes issue(s).

In my custom project I notices that when `lv_task_handler` timing is not adhered to it leads to performance degradation. This PR is more or less a way to raise a possible LVGL misuse.

jakkra:
Thank you, I'll take a look and figure out what the best approach is and if this can improve the performance.

PR #23, opened 2023-07-20
Initial hardware changes to add an external flash

vChavezB:


jakkra:
I have been playing around with what you have trying to avoid changing as much as possible. You have moved some parts a bit causing a bit of DRC error and unconnected stuff so that nees some work. Since I'm lazy I tried fitting it without having to move much other things around and almost got it working, but hard with the big GND pad beow.

Here are a few notes:
- If we could either skip the big GND pad below the chip or find a IC without it, it would be so MUCH easier. 
- Not sure if there are some that don't require this big GND pad below it.
- You have 2 10k resistors as pulldowns, if we skip those and avoid adding new ones, we can connect those traces to the R5 10k resistor that sits very close to the new flash chip anyway.

vChavezB:
Well the pad according to datasheet is not connected. I just added it for extra mechanical support, I could leave it unconnected.

> You have two 10k pulldowns resistors, if we skip one or both, we can connect those traces to the R5 10k resistor for example that sits very close to the new flash chip anyway. This could make it easier also.

I will have a look on what you mean.  If I remember correctly the WP and HOLD pin of the flash are open drain so they require the pull ups.

Yes the DRC, is something I will def. check again. 

When I have produced the PCB and checked the first draft version I will make an update on this PR and also verify if the flash works (firmware).

jakkra:
I upgraded the project to kicad 7 yesterday to review the new drc errors and all of them can be ignored. So let me know after you fixed the erros you think are "real" and I'll review the rest. And setup ignore config for them.

vChavezB:
Updated the pcb, feel free to check it out. This is the version that I produced and currently testing. Keep in mind that its a proof of concept and designed really quick just to test device firmware upgrade functionality. I think the DRC should not complain about the changes I made.

jakkra:
> 
Updated the pcb, feel free to check it out. This is the version that I produced and currently testing. Keep in mind that its a proof of concept and designed really quick just to test device firmware upgrade functionality. I think the DRC should not complain about the changes I made.

Hi, I checked it now and looks nice. Some comments.
- Is R14 needed? Can't we just remove it? Also when flash is NOT used and CS is high, wouldn't there be a current leak due to this pullup?
- I'd like to do some small cleanup of some traces before merging also. Some can be made thicker etc.
- I realized it's super hard to review the routing diff, but from what I can see it looks ok. Will do one more sitting to be sure.

vChavezB:
> wouldn't there be a current leak due to this pulldown

Its a pull up configuration. I think in the case CS pin is high then you would have a voltage difference of the 1.8V - GPIO High (around 1.8V), which I think should give a minimal current leak.

I added the pullup when the device is rebooting, crashes or leaves CS low due to a bug. When CS Low, if the flash is in  deep-power mode (2uA)  then it might wake up and be in standby mode (10uA). Note. I havent measured the real current consumption. 

As I understand from the readme and docs, at the moment the device does not reach the microamps range in sleep mode so it wouldnt really make a difference if the pullup is removed.

There is a tool to check Kicad diff visually

https://github.com/Gasman2014/KiCad-Diff

I will check it out and see if I can make it work.

jakkra:
> Its a pull up configuration. I think in the case CS pin is high then you would have a voltage difference of the 1.8V - GPIO High (around 1.8V), which I think should give a minimal current leak.

My bad, missed that, then it's perfectly fine! Sounds good!

> There is a tool to check Kicad diff visually
> https://github.com/Gasman2014/KiCad-Diff

I'll take a look!

jakkra:
> https://github.com/Gasman2014/KiCad-Diff

Seems to work nicely, super nice tool!
Had to do one patch as it crashed:
image

image
image

PR #26, opened 2023-07-28
Added DFU with MCUBoot and external flash

vChavezB:
- This change adds support for the MCUBootloader with the external flash AT25SL128A. The spi nor flash driver has been activated, configuration for MCUMgr and the BT SMP transport for DFU have been activated.
- Overlay for MCUBoot child image and proper configuration for net core have been also included.
- Enforce use of P0.00 and P0.00 pins as GPIOs in defconfig of board files.
- Declaration of static partition for DFU with use of external flash for LittleFS, NVS, and MCUBoot secondary partition.

I have added support for an external flash based on the design I proposed here

https://github.com/jakkra/ZSWatch/pull/23

DFU works however, I noticed  DFU transfer over BT is slow (0.7 kb/s). I added the appropiate changes to HCI ACL transmission buffers and MTU TX buffer but I am not sure if its related to flash writing speed or another KConfig I am missing for the NRF5340 MCU.


jakkra:
I think the reasoning for slow upload is due to me using like 500ms connection intervals. At the start of the upgrade try calling https://github.com/jakkra/ZSWatch/blob/main/app/src/ble_comm.c#L195

See if that speeds it up.

Awesome that you got this working! Just need to make sure it's backwards compatible and works fine when there is no external flash also, i.e. Using mcuboot at all could be a separate  conf and overlay file maybe?

Also do you know if this works same way when running pure Zephyr and not ncs?

vChavezB:
Yes, I think the bootloader, mcumgr and BT SMP protocol could be a separate conf file. This also allows to separate the build for posix version which does not require mcuboot. 

For mcuboot you would need something like this in cmake

```
set(mcuboot_DTC_OVERLAY_FILE ${OVERLAY_DIR}/flash.overlay)
```

> Also do you know if this works same way when running pure Zephyr and not ncs?

I am not sure about it. The downstream version of nordic uses the [partition manager tool](https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/scripts/partition_manager/partition_manager.html). I think upstream zephyr uses the device tree so the partitions need to be set in device tree. Have not really use mcuboot with upstream zephyr for nordic MCUs so its an area to explore...

Will have a look on the suggestion to speed up the BLE connection. I should also add a timeout in case the DFU is interrupted then I can call `ble_comm_long_connection_interval`.

jakkra:
Sounds good!
Yeah partition manager is not used in upstream. But I assume it's just an entry in some devicetree file, for now this is fine, but I want the code to be compatible with upstream Zephyr. I can also take a look when I have time to see what's needed for it to work in both.

Regarding connection interval, I assume there is some callback maybe we can register on if FW upload fails or times out?

Hmm do you know if you can run Mcuboot with Ota in posix, if so it would be a nice way to experiment with it?

vChavezB:
> Regarding connection interval, I assume there is some callback maybe we can register on if FW upload fails or times out?

Yes there are callbacks for mcumgr when the transfer is in progress. 

https://docs.zephyrproject.org/latest/services/device_mgmt/mcumgr_callbacks.html

In addition there are also callbacks  to the bootloader when an update is in progress (i.e., when the swap algorithm starts moving the partitions), so in case there is enough space in mcuboot then the graphics library could be added in the mcuboot child image to show a small widget that an update is in progress.

Yes to make it work on both zephyr and ncs would be interesting to test. For ncs variant you need the partition.yml static file and to set the property [nordic,pm-ext-flash](https://github.com/jakkra/ZSWatch/blob/37be204c0c9c09029b9f3eb48eabde50afec1939/app/boards/arm/zswatch_nrf5340/zswatch_nrf5340_common.dts#L19).

vChavezB:
Your suggestion worked. Now it takes around  2 minutes for image transfer. (5 to 7 KB/s)

Screenshot_2023-07-29-23-05-07-765-edit_no nordicsemi android mcp

vChavezB:
BTW I forgot to add the static partition file, it should now be on the branch and the dfu callback changes for faster image transfer.

vChavezB:
I was trying to compile with mcuboot but I think its not possible.

See the Kconfig option [CONFIG_USED_DT_CODE_PARTITION](https://docs.zephyrproject.org/3.4.0/kconfig.html#CONFIG_USE_DT_CODE_PARTITION) which is selected by default when [CONFIG_BOOTLOADER_MCUBOOT](https://docs.zephyrproject.org/3.4.0/kconfig.html#CONFIG_BOOTLOADER_MCUBOOT) is selected. 

`CONFIG_USED_DT_CODE_PARTITION` depends on `CONFIG_HAS_FLASH_LOAD_OFFSET` but if you see its [dependencies](https://docs.zephyrproject.org/3.4.0/kconfig.html#CONFIG_HAS_FLASH_LOAD_OFFSET) it does not target posix.

What I think perhaps (not sure) can be tested in posix is the Simple Management Protocol over Bluetooth. As posix allows to connect to an hci controller in linux you can (maybe?) connect to the mcumgr cli or nrf connect app and try to upload an image. I am not sure if this would work as mcumgr verifies the header and metadata of the image, but since mcuboot is not supported for posix you could not generate an updatable image (.bin) that you could send over the mcumgr cli /nrf connect app. 

Just some thoughts/theories on this.

jakkra:
> Your suggestion worked. Now it takes around 2 minutes for image transfer. (5 to 7 KB/s)

Nice, still sounds a bit slow? But I have not used it before so I don't know what to expect.

> What I think perhaps (not sure) can be tested in posix is the Simple Management Protocol over Bluetooth. As posix allows to connect to an hci controller in linux you can (maybe?) connect to the mcumgr cli or nrf connect app and try to upload an image. I am not sure if this would work as mcumgr verifies the header and metadata of the image. In addition, mcuboot is not supported for posix so you cant send a valid image for mcumgr to decode.
Just some thoughts/theories on this.
If this is the case then in the posix overlay/conf, the bootloader should be disabled.

Thanks for look it up, as you say it seems not fully supported. Was just an idea, but let's skip that then.

I built you branch, some kconfig warning but they seems to be mostly due to me putting BT stuff in the default net core conf. Also moved some config that I think was in the wrong conf file. See my patch file and see if it still works for you after this:


[patch.txt](https://github.com/jakkra/ZSWatch/files/12208140/patch.txt)

vChavezB:
The patch does not break the changes I introduced. I found out that the MCUMgr already has an automatic way of changing connection parameters when SMP is active.

https://github.com/zephyrproject-rtos/zephyr/blob/9dc531be481a2e67da5c2875339b4386e93e823c/subsys/mgmt/mcumgr/transport/src/smp_bt.c#L37-L50

Here are the Kconfig options

https://docs.zephyrproject.org/latest/kconfig.html#!CONFIG_MCUMGR_TRANSPORT_BT_CONN_PARAM_CONTROL


The default options when mcumgr is active and SMP commands are sent is 7.2ms, timeout latency , which give around 14.5 Kb/s. Which I think is around ~1 minute for uploading a 1MB image.

Basically Mcumgr changes back to the user preferred settings. These are the ones I set trying to match your settings in `ble_comm_long_connection_interval`. I could not match them correctly as there are some static asserts that prevent it me to do it.

https://github.com/zephyrproject-rtos/zephyr/blob/9dc531be481a2e67da5c2875339b4386e93e823c/subsys/bluetooth/host/gatt.c#L184

```
CONFIG_BT_PERIPHERAL_PREF_MIN_INT=400
CONFIG_BT_PERIPHERAL_PREF_MAX_INT=500
CONFIG_BT_PERIPHERAL_PREF_LATENCY=0
CONFIG_BT_PERIPHERAL_PREF_TIMEOUT=1000
```

jakkra:
Perfect, then we can also update ble_comm.c to use the same kconfig so that to change them only the prj.conf needs update maybe.

JeffWelder:
Hi there, just chiming in with a Kconfig I use for nRF53 to get MCUmgr speed back up found in some Nordic samples (samples/bluetooth/peripheral_status/boards/thingy53_nrf5340_cpuapp.conf):
```CONFIG_MCUMGR_TRANSPORT_BT_CONN_PARAM_CONTROL=y
# Enable MCUmgr Packet Reassembly feature over Bluetooth and its
# configuration dependencies. MCUmgr buffer size is optimized to fit one SMP
# packet divided into five Bluetooth Write Commands, transmitted with the
# maximum possible MTU value: 498 bytes.
CONFIG_MCUMGR_TRANSPORT_BT_REASSEMBLY=y
CONFIG_MCUMGR_TRANSPORT_NETBUF_SIZE=2475
CONFIG_MCUMGR_GRP_OS_MCUMGR_PARAMS=y
# Increase system workqueue stack size to prevent stack overflows while handling
# SMP requests.
CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=4096

CONFIG_BT_L2CAP_TX_MTU=498
CONFIG_BT_BUF_ACL_TX_SIZE=502
CONFIG_BT_BUF_ACL_RX_SIZE=502
```

vChavezB:
Thanks for the suggestion @JeffWelder . Actually the options you mentioned are already applied.

See

- https://github.com/jakkra/ZSWatch/blob/8847bc609ee9cfcb92c1c033027a22b230cf086a/app/prj.conf#L3
- https://github.com/jakkra/ZSWatch/blob/8847bc609ee9cfcb92c1c033027a22b230cf086a/app/prj.conf#L220-L223
- https://github.com/jakkra/ZSWatch/blob/8847bc609ee9cfcb92c1c033027a22b230cf086a/app/prj.conf#L213-L214

JeffWelder:
Ahh, great!  I must have searched the prj.conf for the older ncs symbol `CONFIG_MCUMGR_SMP_BT_CONN_PARAM_CONTROL`.

You guys are doing great work, and I'm hopefully going to assemble a watch in the future. 😄

vChavezB:
> Perfect, then we can also update ble_comm.c to use the same kconfig so that to change them only the prj.conf needs update maybe.

I have added the changes.

vChavezB:
I left the [dfu.c](url) source file but its not needed at the moment, unless you see a use case to know when an image event happens

https://docs.zephyrproject.org/3.4.0/services/device_mgmt/mcumgr_callbacks.html#c.img_mgmt_group_events

The other point is , should we reject DFU unless user allows it? I.e., user enters update mode. See the comment I made here

https://github.com/jakkra/ZSWatch/blob/76fa2ef141ccffb6feb440e10cce9bc6e0c8b3c7/app/src/dfu.c#L22

jakkra:
> I left the dfu.c source file but its not needed at the moment, unless you see a use case to know when an image event happens

Let's keep dfu.c, we could do some nice progress bar or something on the watch to show status etc. leter.

> The other point is , should we reject DFU unless user allows it? I.e., user enters update mode. See the comment I made here

I think as long as the remote device is bonded it's fine. We could add a setting in the future to enable it, but for now I think this is just fine.
Maybe use https://docs.zephyrproject.org/latest/connectivity/bluetooth/api/connection_mgmt.html#c.bt_conn_get_security and check if the level is >= BT_SECURITY_L2 (encryption). The only way to get the link encrypted is to bond, and this needs to be allowed in the watch settings. Hence we trust this remot device.

vChavezB:
Ah ok, that seems a good idea. For this api can I pass a NULL and get the first connection or do I need to have a pointer to ` struct bt_conn` ?

jakkra:
> Ah ok, that seems a good idea. For this api can I pass a NULL and get the first connection or do I need to have a pointer to ` struct bt_conn` ?

Not sure, don't you have the connection in the mcumgr callbacks?

jakkra:
Try null otherwise the current conn is here and you can add a get function here https://github.com/jakkra/ZSWatch/blob/13c65f447cda68df36aa47224016e859abb3661f/app/src/ble_comm.c#L67C26-L67C26

vChavezB:
Ah btw, I found out that the NRF5340 DK has a 8 MB flash. If you have one I could add an overlay so you can test the DFU changes with the NRF Connect app.

jakkra:
Yeah I also thought I'd add that later, I have one.

vChavezB:
I made an attempt here

https://github.com/vChavezB/ZSWatch/tree/devkit_flash


The flash driver works but the firmware resets when main is called by zephyr. Im not sure why...

jakkra:
Don't worry about it, I'll give it a shot later. For now let's merge your PRs if you are ready. I will just fix the kicad warnings in your other PR and both can be merged then.

vChavezB:
Ok sounds good to me!

vChavezB:
Before I forget I will add a overlay and conf to skip mcuboot otherwise it wont compile for posix

jakkra:
> Before I forget I will add a overlay and conf to skip mcuboot otherwise it wont compile for posix

Perfect, thanks!

jakkra:
> I made an attempt here
> 
> https://github.com/vChavezB/ZSWatch/tree/devkit_flash
> 
> The flash driver works but the firmware resets when main is called by zephyr. Im not sure why...

I just compiled your branch, and uploaded a new FW from nRF Connect app and it worked fine. However Bluetooth stopped working after updating the new image. I'm going to re-try again.

vChavezB:
Did you bridge the VDD, CS, SI, SO pins of the flash? The other two pins should not be bridged.

jakkra:
> Did you bridge the VDD, CS, SI, SO pins of the flash? The other two pins should not be bridged.

TBH I ran it on a u-blox NORA-W10 EVK (with no modifications), but it should be pin compatible with the nRF5340dk. From the log at least is looked like the SPI flash communication worked. And I could do DFU once sucessfully, I assume if flash didn't work then I would not get this far?

Anyway it seems as my Bluetooth issue is not with DFU, if I just run "Flash" and let nRF connect plugin flash all images in VSCode I get initialize errors for Bluetooth. 
I guess it's some kconfig that broke when compiling for nrf5340dk which caused this as it works for you when flashing on the actual watch.

vChavezB:
Initialize errors for bluetooth happen when the flash doesnt work. Since `bt_enable` tries to load NVS settings (i.e., flash).

From the ublox DK it seems you also have to bridge the flash pins.

grafik.

You can put a breakpoint in  zephy/drivers/flash/spi_nor.c and check if the driver is initialized correctly as well.

jakkra:
> Initialize errors for bluetooth happen when the flash doesnt work. Since `bt_enable` tries to load NVS settings (i.e., flash).
> 
> From the ublox DK it seems you also have to bridge the flash pins.
> 
> grafik.
> 
> You can put a breakpoint in `zephyr/drivers/flash/spi_nor.c` and check if the driver is initialized correctly as well.

Ah probably is that, will try tonight again with this. Thanks!

vChavezB:
I have added conditional compilation for DFU when the board is not posix. This means that old watches are not compatible with this firmware.

As a workaround I suggest instead that another board is created as revision 2 (with the flash device tree changes). Then DFU will be compiled only for this new revision. It can also be decided to make the dfu configs optional so the user will have to set this via west. I like better the option of just selecting the board. 

Just some thoughts.

jakkra:
Can you do that without copying the whole boards/arm/zswatch_nrf5340, because I really don't want to duplicate that one?

jakkra:
Yes I think so. So you mean something like this:
boards/zswatch_nrf5340/<like today>
boards/zswatch_v1_nrf5340/zswatch_v1_nrf5340.dts => includes zswatch_nrf5340 somehow
boards/zswatch_v2_nrf5340/zswatch_v2_nrf5340.dts => includes zswatch_nrf5340 somehow

vChavezB:
yes something like that. When I have the chance I will try to test this idea out and maybe add a github actions to compile for each revision as a sanity check.

jakkra:
> yes something like that. When I have the chance I will try to test this idea out and maybe add a github actions to compile for each revision as a sanity check.

Sound awesome!
I'm not sure how to do that, but I can ask on Zephyr Discord to see how to do this. or if we can get some pointers.

jakkra:
I think this is what we want https://docs.zephyrproject.org/latest/hardware/porting/board_porting.html#number-revision-matching

vChavezB:
Looks promising. Also the CmakeList.txt source files have to be defined manually. Currently you have global recursion to include all *.c files. The dfu.c should only be compiled if the board is not posix.

vChavezB:
The board revision works after your update. Something like this could be added. will have to test it:

```
string(REGEX MATCH "zswatch_nrf5340_cpuapp(_ns)?@([0-9]+)$" ZSWATCH_BOARD "${BOARD}")
if(ZSWATCH_BOARD )
    set(ZSWATCH_REV ${CMAKE_MATCH_2})
    if(${ZSWATCH_REV} GREATER_EQUAL 2)
        set(DFU_BUILD true)
        set(PM_STATIC_YML_FILE ${CMAKE_CURRENT_SOURCE_DIR}/partition.yml)
        set(OVERLAY_CONFIG ${CMAKE_CURRENT_SOURCE_DIR}/boards/dfu.conf)
        set(mcuboot_DTC_OVERLAY_FILE ${CMAKE_CURRENT_SOURCE_DIR}/overlays/mcuboot.overlay)       
    endif()
endif()
```

vChavezB:
I have tested the changes and seems to work, also modified the CI/CD to build the different revisions.

jakkra:
Great! I cannot merge due to conflicts in .github/workflows/build.yml
Not at my computer, but can fix it later, is a simple one.

jakkra:
Well, I think I messed up a bit, I took my standard apparoch to rebase on main in order to solve the conflict, but due to the merges we have done, it got messed up without me realising. Anyway it's merged now and thanks so much for the contribution @vChavezB !

PR #32, opened 2023-08-01
Use board revisions to handle external flash.

jakkra:
Seems to work as expected with board revisions.
image

Or in terminal: `west build -p -b zswatch_nrf5340_cpuapp@<rev_nbr>`

Had to add dummy files for rev1, and there had to be rev files for net core also, otherwise it wouldn't compile.

Not super happy about the` if(${BOARD} STREQUAL "zswatch_nrf5340_cpuapp@2")` part, but maybe fine for now, what do you say?


jakkra:
Github fighting with me should go into a fork of this repo... 
Fixed now https://github.com/vChavezB/ZSWatch/pull/4

PR #33, opened 2023-08-08
ci: build from outside the container

coderbyheart:
This fixes the build by not running the action inside the container but using the Docker image in the action.


jakkra:
Thank you very much, works like a charm!

PR #34, opened 2023-08-11
Update GETTING_STARTED

AlmazShai:
Following steps described in getting started guide I found some step unclear and decided to fix that. 
I also found that application can be executed on native_posix without using any external dongle, but this applicable only if you have build-in BLE module. I tested both variants native posix and native posix + dongle and updated documentation based on my experience.
Please review carefully and give your feedback.


jakkra:
Very nice improvement! Thanks!
Will try the native posix without dongle also to see if I can get that up and running too.

PR #37, opened 2023-08-15
Rework popup notification

AlmazShai:
Reworked notification UI to make full message visible on screen. Notification message size now is limited to 3 lines of message, but the user can expand the message by clicking on it. Clicking on expanded message one more time will remove the message.

I like to replace "Bell" symbol with small application symbol in future, but I would like to take I look to notification manages to be able to understand if it's possible to show sender name instead of application name.


AlmazShai:
Short form of notification message now looks like:
photo_2023-08-15_20-45-38

app/src/ui/zsw_popup_notifcation.c
```diff
@@ -1,108 +1,153 @@
...
+    // TODO: replace with application symbol
+    strcpy(title_text, LV_SYMBOL_BELL);
+    strcat(title_text, " ");
+    strcat(title_text, title);
+
+    // create title text
+    notif_box.title = lv_label_create(notif_box.panel);
+    lv_obj_set_x(notif_box.title, -5);
+    lv_obj_set_y(notif_box.title, -10);
+    lv_label_set_text(notif_box.title, title_text);
```
jakkra, line 59: You can use `lv_label_set_text_fmt(notif_box.title, "%s %s", LV_SYMBOL_BELL, title);` and skip the `title_text` and strcpy lines above. `lv_label_set_text_fmt` is a very handy function.

app/src/ui/zsw_popup_notifcation.c
```diff
@@ -1,108 +1,153 @@
...
+    lv_obj_clear_flag(notif_box.panel, LV_OBJ_FLAG_SCROLLABLE);
+    lv_obj_set_style_radius(notif_box.panel, 15, 0);
+    lv_obj_set_style_bg_color(notif_box.panel, lv_color_hex(0x0C0C1A), 0);
+    lv_obj_set_style_border_color(notif_box.panel, lv_palette_main(LV_PALETTE_BLUE_GREY), 0);
+    lv_obj_set_style_border_width(notif_box.panel, 1, 0);
+    lv_obj_add_event_cb(notif_box.panel, on_notification_expand, LV_EVENT_PRESSED, NULL);
+
+    // Append symbol before title
+    char title_text[strlen(title) + strlen(LV_SYMBOL_BELL) + 2];
+    // TODO: replace with application symbol
```
jakkra, line 50: I think we can do it this way later in order to easily add some more notification icons. https://docs.lvgl.io/8.3/overview/font.html#add-new-symbols Those will also tacke much less flash than the icons I have today.

app/src/ui/zsw_popup_notifcation.c
```diff
@@ -1,108 +1,153 @@
...
+static void remove_notification(uint32_t notif_id);
+
+static on_close_notif_cb_t on_close_cb;
+static uint32_t active_notif_id;
 static lv_timer_t *auto_close_timer;
+static notif_box_t notif_box;
 
 LV_IMG_DECLARE(messenger);
 LV_IMG_DECLARE(gmail);
 LV_IMG_DECLARE(notification);
```
jakkra, line 24: You can remove those and also delete the .c files in images folder.

jakkra (changes requested):
Tested and works like a charm! Just some minor things.

app/src/ui/zsw_popup_notifcation.c
AlmazShai: Fixed in last commit

PR #39, opened 2023-08-19
Base64 encode implemented for notifications

AlmazShai:
I found that sometime "sender" or "body" field encoded in base64 format, it mostly depends on application.
I also tried to add multilanguage support but find that it's not supported on mobile application side. Maybe sometime they fix this and we also implement dome multilanguage fronts. These changes are basically the first step toward it, since all multilanguage seems to be encoded in base64 format.


jakkra:
Interesting, haven't stumbled upon this before.

jakkra:
So you mean if a none English character is received in sender for example, gadgetbridge does not send to the watch those letters correctly?

AlmazShai:
In fact, then sending non ASCII characters via gadgetbridge debug feature it send "???" symbols in base64 format, I guess that Bangle.js does not support non-ASCII symbols. 
And some applications like "Telegram" send sender name and body in Base64 format, in case then this fields contains only ASCII character ZSWatch can parse them correctly now (before it convert to empty fields).

jakkra:
I seem to be getting Swedish letters in UTF-8
Å - rendered as \xc5 for example

So that at least should be fixable, maybe for me just using a font with those characters. Not sure how lvgl handles it.

But this is me sending from the debug field in gadgetbridge, maybe different then the source is an app.

PR #91, opened 2023-10-17
Feature rework input subsystem

jakkra:


Kampi:
What is the issue? Do we need a modification of the device tree?

jakkra:
> What is the issue? Do we need a modification of the device tree?

I'd assume so as the input changes are right now only in v3 devicetree. I'll fix this today. Want to get this merged so I can continue on stuff based on this.

jakkra:
> > > What is the issue? Do we need a modification of the device tree?
> > 
> > 
> > I'd assume so as the input changes are right now only in v3 devicetree. I'll fix this today. Want to get this merged so I can continue on stuff based on this.
> 
> The longpress device is missing in the device trees for revision 1 and 2. You can also check the pinout for the buttons and make sure they are correct. The touch controller is available in rev 1 to 3.

Yep will take a look, not much change needed :)

PR #98, opened 2023-10-30
Sensors and project improvements

jakkra:


.vscode/launch.json
```diff
@@ -0,0 +1,24 @@
...
+  "configurations": [
+    {
+      "name": "C/C++ Runner: Debug Session",
+      "type": "cppdbg",
+      "request": "launch",
+      "args": [],
+      "stopAtEntry": false,
+      "externalConsole": true,
+      "cwd": "f:/Git/ZSWatch/app",
+      "program": "f:/Git/ZSWatch/app/build/Debug/outDebug",
```
jakkra, line 12: This is for Native Posix right? If so replace paths with "${workspaceFolder}/build/zephyr/zephyr.exe" => {workspaceFolder}/ instead of hardcode paths

.vscode/settings.json
```diff
@@ -1,13 +1,70 @@
...
+    "${workspaceFolder}\\app"
+  ],
+  "nrf-connect.boardRoots": [
+    "${workspaceFolder}/app",
+    "${workspaceFolder}\\app"
+  ],
+  "nrf-connect.debugging.bindings": {
+    "${workspaceFolder}/app/build": "Launch build"
+  },
+  "C_Cpp_Runner.cCompilerPath": "gcc",
```
jakkra, line 13: I don't think this should be needed? All of this should be handled by nRF COnnect VSCode plugin already?

PR #101, opened 2023-11-02
Notification fixes

jakkra:


app/src/main.c
```diff
@@ -124,13 +123,6 @@ static void run_input_work(struct k_work *item)
 
     // Don't process the press if it caused wakeup.
     if (zsw_power_manager_reset_idle_timout()) {
-        // NOTE: As soon as the touch controller wakes up the MCU through an interrupt the event is also
-        // passed into LVGL (interrupt -> input subsystem -> LVGL). We don´t want any LVGL action after
-        // a touch event. Also make sure that a wake up through a button doesn´t block the first input of the touch screen.
-        if (container->event.code > INPUT_KEY_4) {
-            is_woken_by_touch = true;
```
jakkra, line 131: @Kampi Do you remember if this had some use and it got lost somehow, because right now it's doing nothing?

Kampi: I add it to prevent a double action after waking up the watch. The code was added in an early stage and based on existing code. Maybe it can get dropped now. 

In the beginning there was an issue where waking up the watch also triggers an action through this button (pressing it result in waking up and opening the menu). This was not the behavior that we wanted, because we wanted to have two presses for this. Maybe it's obsolete now.

jakkra: Ok, I think I sometimes have noticed this, maybe needs refixing, will try to take notice more if it happens

PR #107, opened 2023-11-14
Migrate to NCS SDK v2.5.0

ldab:
Build using NCS SDK v2.5.0, needed for some iOS stuff planned on the next PR:

1. Changed the `west.yml` remote to match nrf sdk
2. removed the `app/zephyr` folder that caused some death loop dependency (???)
3. Added a partition file for the static memory
4. ~~removed error handler from main as it is defined on `nrf/lib/fatal_error/fatal_error.c`~~

I haven't tested on the HW, maybe in few days :wink:


app/src/main.c
```diff
@@ -605,15 +605,3 @@ static void on_zbus_ble_data_callback(const struct zbus_channel *chan)
     }
 }
 
-void k_sys_fatal_error_handler(unsigned int reason, const z_arch_esf_t *esf)
```
jakkra, line 608: Don't remove as it's good to have when doing a zephyr only build.
Wrap it in:
#ifdef  CONFIG_RESET_ON_FATAL_ERROR

app/zephyr/module.yml
```diff
@@ -1,20 +0,0 @@
-# Copyright (c) 2021 Nordic Semiconductor ASA
-# SPDX-License-Identifier: Apache-2.0
-
-build:
```
jakkra, line 4: Will try it out to make sure this doesn't break boards discovery etc.

jakkra (commented):
I'm going to look around if there is a good way to have both ncs and pure zephyr support. For example multiple west.yml or similar.

app/CMakeLists.txt
```diff
@@ -10,6 +10,8 @@ if(ZSWATCH_BOARD)
         set(PM_STATIC_YML_FILE ${CMAKE_CURRENT_SOURCE_DIR}/partition.yml)
         set(OVERLAY_CONFIG ${CMAKE_CURRENT_SOURCE_DIR}/boards/dfu.conf)
         set(mcuboot_DTC_OVERLAY_FILE ${CMAKE_CURRENT_SOURCE_DIR}/overlays/mcuboot.overlay)
+    else()
+        set(PM_STATIC_YML_FILE ${CMAKE_CURRENT_SOURCE_DIR}/partition_zs.yml)
```
jakkra, line 14: Hmm,  I'm not sure how to use different pm.yml depending on board, hence build failing. Currently it's handled cleanly in the dts. May need to look this up.

jakkra: Here it is: [CONFIG_RESET_ON_FATAL_ERROR ](https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/scripts/partition_manager/partition_manager.html#configuring-static-partitions)

ldab: They are all the same as far as I can tell, have added to the Cmake to include the other ones

jakkra: Yes for rev3 they are the same, but not for rev 1 (doesn't have them) but it may still compile as partition manager works differently and then it's all good. CI will tell :)

ldab: `pm_static_<board>_<revision>.yml` will need 6 files bevause of `release` and `debug`

I added 2x to Cmake, let me know what you think

jakkra: Seems to compile for rev1 also with just one pm.yml, so we can keep it like this. I don't think a specific on is needed for the 5340_dk or?

jakkra:
Tested it on real HW and I get 
image
I guess it's just the settings missing from pm file, but not sure yet.

app/CMakeLists.txt
ldab: If the DK has the ‘external_partition’ area it does not build

jakkra: Ah didn't see any difference between the files, now I saw it. The dk does have external flash also, so need to find out why it's not working, or maybe named something other than "external_flash" But I have asked nordic if there is a way to disable the partition manager, so let's see

ldab:
> Tested it on real HW and I get image I guess it's just the settings missing from pm file, but not sure yet.

I missed the settings partition, should be good now

app/CMakeLists.txt
ldab: It didn’t work on the DK because the external flash overlay was not set, all good now

jakkra:
> > Tested it on real HW and I get image I guess it's just the settings missing from pm file, but not sure yet.
> 
> I missed the settings partition, should be good now

Still crashing same issue. I'm debugging it atm.

jakkra:
Fixed:
This commit to not override setting partititon for Zephyr https://github.com/jakkra/ZSWatch/commit/02c806c4b0455d2c50e0c7456c2f981876278973
Plus the newly pushed stuff in this branch.

jakkra:
While it's all working I found there is an issue with scrolling, every time you release the display, it will also press. @Kampi I have had this issue before, but seems as it's back now. You recently remade some of the input touch code, any idea?

Kampi:
@jakkra is this issue new and related to 2.5 or does it also exist in the current version? I haven't seen this issue on my watch so far, so I don't have any idea. If it's new to 2.5 I have to do the upgrade too and investigate this issue.

jakkra:
> @jakkra is this issue new and related to 2.5 or does it also exist in the current version? I haven't seen this issue on my watch so far, so I don't have any idea. If it's new to 2.5 I have to do the upgrade too and investigate this issue. 

It's new and related to 2.5 upgrade. I guess something changed.

Kampi:
> > @jakkra is this issue new and related to 2.5 or does it also exist in the current version? I haven't seen this issue on my watch so far, so I don't have any idea. If it's new to 2.5 I have to do the upgrade too and investigate this issue.
> 
> It's new and related to 2.5 upgrade. I guess something changed.

Okay then I (we) have to take a look at it.

jakkra:
> > > @jakkra is this issue new and related to 2.5 or does it also exist in the current version? I haven't seen this issue on my watch so far, so I don't have any idea. If it's new to 2.5 I have to do the upgrade too and investigate this issue.
> > 
> > It's new and related to 2.5 upgrade. I guess something changed.
> 
> Okay then I (we) have to take a look at it. 

Yup, just wanted to know if yoy had any quick idea what it could be. @ldab let me know if you want to look at it, otherwise I can take a look.

ldab:
> > > > @jakkra is this issue new and related to 2.5 or does it also exist in the current version? I haven't seen this issue on my watch so far, so I don't have any idea. If it's new to 2.5 I have to do the upgrade too and investigate this issue.
> > > 
> > > 
> > > It's new and related to 2.5 upgrade. I guess something changed.
> > 
> > 
> > Okay then I (we) have to take a look at it.
> 
> Yup, just wanted to know if yoy had any quick idea what it could be. @ldab let me know if you want to look at it, otherwise I can take a look.

If yous have an idea where the problem might be I’d leave it to you due to your familiarity with the thing. Otherwise I can look at it, I think I experienced on posix also, it was quite tricky to scroll up without opening another menu

jakkra:
> > > > > @jakkra is this issue new and related to 2.5 or does it also exist in the current version? I haven't seen this issue on my watch so far, so I don't have any idea. If it's new to 2.5 I have to do the upgrade too and investigate this issue.
> > > > 
> > > > 
> > > > It's new and related to 2.5 upgrade. I guess something changed.
> > > 
> > > 
> > > Okay then I (we) have to take a look at it.
> > 
> > Yup, just wanted to know if yoy had any quick idea what it could be. @ldab let me know if you want to look at it, otherwise I can take a look.
> 
> If yous have an idea where the problem might be I’d leave it to you due to your familiarity with the thing. Otherwise I can look at it, I think I experienced on posix also, it was quite tricky to scroll up without opening another menu

Ah nice, it's reproduceable in posix, then this will be easier to fix. The touch stuff does then not go through ZSWatch code at all, hence looks like a pure Zephyr issue. And probably reproduceable with lvgl samples. Will try.

Kampi:
That´s helpful. If this issue also exists in Posix then we don´t have an issue with the touch driver in combination with LVGL. So it looks like a pure LVGL / display handling issue for me. Scrolling is handled by LVGL. My suggestion is to start here

https://github.com/jakkra/ZSWatch/blob/02c806c4b0455d2c50e0c7456c2f981876278973/app/src/main.c#L492

jakkra:
TODO: Make sure this is part of ncs 2.5 https://github.com/zephyrproject-rtos/zephyr/pull/62756
That fixed exactly this problem before. If it's not part of it we can make a patch for it, or instead of using ncs-2.5.0 in west.yml we can try with "main" which get latest unreleased ncs

ldab:
> TODO: Make sure this is part of ncs 2.5 [zephyrproject-rtos/zephyr#62756](https://github.com/zephyrproject-rtos/zephyr/pull/62756) That fixed exactly this problem before. If it's not part of it we can make a patch for it, or instead of using ncs-2.5.0 in west.yml we can try with "main" which get latest unreleased ncs

https://github.com/jakkra/ZSWatch/pull/107/commits/b75b44dce6f7a50c327ddc8e21f71a1a10ee9fa2 applies the patch, the problem is fixed on my side with POSIX

jakkra:
Ok, I'll merge this then and we run ZSWatch on NCS (at least for a while) as it simplifies alot with the iOS integration which requires a bunch of code from NCS.

Issue #44, opened 2023-08-21
Auto patches to Zephyr when compiling


jakkra:
There are some patches, for example to upscale native_posix SDL display. 
Would be handy to automatically apply those patches when complinging.

vChavezB:
Golioth has a script to run a patch command when using the manifest file

https://github.com/golioth/golioth-zephyr-sdk/blob/main/scripts/west_commands/patch.py

and you can see it in action here

https://github.com/golioth/golioth-zephyr-sdk/blob/cc146c1e0d4f2c666a37da3b6de7cc5902367484/west-ncs.yml#L15

This issue would be related to first setting up a manifest file as mentioned in https://github.com/jakkra/ZSWatch/issues/35

ldab:
@jakkra I can only see “ app/zephyr_patches/sdl_upscale.patch” is there another one?

Kampi:
@ldab we have a patch for the nPM1300 driver which is necessary for later revisions with the new power management. It´s not checked in in Git, so you can add it if you want

[ntc_fix.patch](https://github.com/jakkra/ZSWatch/files/13367904/ntc_fix.patch)

jakkra:
@ldab do note that this sdl patch does not work anymore, that code was refactored. It's not necessary, so it can be removed.

ldab:
> @ldab do note that this sdl patch does not work anymore, that code was refactored. It's not necessary, so it can be removed.

both `kscan_sdl.c` and `display_sdl.c` are not required anymore? How to make the screen bigger?

jakkra:
> > @ldab do note that this sdl patch does not work anymore, that code was refactored. It's not necessary, so it can be removed.
> 
> both `kscan_sdl.c` and `display_sdl.c` are not required anymore? How to make the screen bigger?

Well they are needed to scale it up, but requires re-doing this patch. But current default not scaled is useable also.

Kampi:
@jakkra @ldab 
CI/CD is reporting this

![image](https://github.com/jakkra/ZSWatch/assets/3458587/7700814c-f032-4d02-8a65-e3dbf02174a3)

jakkra:
> @jakkra @ldab CI/CD is reporting this
> 
> ![image](https://user-images.githubusercontent.com/3458587/285316113-7700814c-f032-4d02-8a65-e3dbf02174a3.png)

I got a fix incoming.

ldab:
Strange, main gives me:

```
-- west build: generating a build system
[...]

PR #108, opened 2023-11-15
Apply patches recursively before build

ldab:
Apply all patches on the `patches/` dir.

Please note that the `display_sdl.c` is still there as I do not know how to make the posix screen bigger, this patch makes it bigger but the watch screen does not resize, just the outer one.

In any case, this PR is more about the `app/CMakeLists.txt`

#44


app/patches/sdl_upscale.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/drivers/display/display_sdl.c b/drivers/display/display_sdl.c
+index 2a457249ed..ed624bb2c0 100644
+--- a/drivers/display/display_sdl.c
++++ b/drivers/display/display_sdl.c
+@@ -54,7 +54,7 @@ static int sdl_display_init(const struct device *dev)
+ #endif /* SDL_DISPLAY_DEFAULT_PIXEL_FORMAT */
+ 		;
+ 
+-	int rc = sdl_display_init_bottom(config->height, config->width, &disp_data->window,
++	int rc = sdl_display_init_bottom(config->height * 4, config->width * 4, &disp_data->window,
```
jakkra, line 10: I don't think touch input works when doing this, at least before. i.e. touch x,y need to be scaled accordingly too? If so it's better to remove this completely and fix it later instead.

ldab: Yes, it does not work, I just left because when writing the cmake I didn’t have any patch

ldab:
When it builds you will see

```
patching file drivers/sensor/npm1300_charger/npm1300_charger.c
Reversed (or previously applied) patch detected!  Assume -R? [n] 
Apply anyway? [n] 
Skipping patch.
1 out of 1 hunk ignored
```

PR #111, opened 2023-11-16
ios Apple Media Service

ldab:
Adds iOS media control.

Few things to note:
1. Apple notification service is still work in progress
2. This is for music control and music info
3. The CI builds for iOS but does not upload the image, this is intentional as I plan to upload when 1 is implemented

image


.vscode/settings.json
Kampi, line 1: You can ignore the settings file

ldab: that was intentional, i thought it could be good to set the default editor

Kampi: There is already a west command to apply the format. @jakkra shall we add this to the settings?

jakkra: I think I also run the same, so as long as it still works when user does not have in installed it's all good.

jakkra: @Kampi as this formats code on save and I use it also, I think it makes sense to add it. And if anyone doesn't like it formatting on save, then just don't install the plugin.

app/src/applications/music_control/music_control_app.c
```diff
@@ -78,6 +84,28 @@ static void on_music_ui_evt_music(music_control_ui_evt_type_t evt_type)
     if (msg_len > 0) {
         ble_comm_send(buf, msg_len);
     }
+
+#elif defined(CONFIG_BLE_USES_AMS)
```
jakkra, line 88: This will do good. Seeing this made me create this ticket to better handle this later https://github.com/jakkra/ZSWatch/issues/112

jakkra (approved):
Awesome!

Kampi (approved):

PR #120, opened 2023-11-19
Fix AMS fail when discover is busy

ldab:
DM discover only allow one instance at the time, on iOS we discover for:

- CTS
- AMS
- ANCS

Thus implement some retrial.


jakkra:
Seems as I broke CI yesterday, let me fix it and then you can rebase in main.

jakkra (approved):
Looks good, just rebase on main and CI should pass.

PR #121, opened 2023-11-19
Current Time Service (CTS) client

ldab:
This may also work for Android, I don't know, haven't tested

It seems like the clock is adjusting for DST, should I pass one hour less?


app/src/ble/ble_cts.c
```diff
@@ -0,0 +1,236 @@
...
+#include <zephyr/zbus/zbus.h>
+
+#include <bluetooth/gatt_dm.h>
+#include <bluetooth/services/cts_client.h>
+
+#include "ble/ble_comm.h"
+#include "ble/ble_cts.h"
+#include "events/ble_data_event.h"
+
+LOG_MODULE_REGISTER(ble_cts, LOG_LEVEL_INF);
```
jakkra, line 28: Probbaly want LOG_LEVEL_WRN as default here to not spam with logs. But up to you for now if you want to keep or not.

app/src/ble/ble_cts.c
```diff
@@ -0,0 +1,236 @@
...
+static void connected(struct bt_conn *conn, uint8_t err)
+{
+    if (err) {
+        LOG_ERR("Connection failed (err 0x%02x)", err);
+        return;
+    }
+
+    has_cts = false;
+
+    err = bt_gatt_dm_start(conn, BT_UUID_CTS, &discover_cb, NULL);
```
jakkra, line 210: Need retries here also as in the other services?

jakkra (approved):
Looks good, just rebase on main and CI should pass.

jakkra:
As the timezone is currently hadrcoded here: https://github.com/jakkra/ZSWatch/blob/main/app/src/zsw_clock.c#L42
And I assume the time you get from iOS is already corrected we have this issue.
Gadgetbridge time is not adjusted.
So some solution is needed here to make both work.

app/src/ble/ble_cts.c
ldab: This happens quite fast and didn’t fail, but it could, I have retried now

ldab:
> As the timezone is currently hadrcoded here: https://github.com/jakkra/ZSWatch/blob/main/app/src/zsw_clock.c#L42 And I assume the time you get from iOS is already corrected we have this issue. Gadgetbridge time is not adjusted. So some solution is needed here to make both work.

Having the timezone hardcoded seems not appropriate, don’t you get it from the app?

jakkra:
> > As the timezone is currently hadrcoded here: https://github.com/jakkra/ZSWatch/blob/main/app/src/zsw_clock.c#L42 And I assume the time you get from iOS is already corrected we have this issue. Gadgetbridge time is not adjusted. So some solution is needed here to make both work.
> 
> Having the timezone hardcoded seems not appropriate, don’t you get it from the app?

Yep, it's just TODO
But seems to be there:
image

ldab:
> > > As the timezone is currently hadrcoded here: https://github.com/jakkra/ZSWatch/blob/main/app/src/zsw_clock.c#L42 And I assume the time you get from iOS is already corrected we have this issue. Gadgetbridge time is not adjusted. So some solution is needed here to make both work.
> > 
> > 
> > Having the timezone hardcoded seems not appropriate, don’t you get it from the app?
> 
> Yep, it's just TODO But seems to be there: image

so it should be ok to merge this PR, it will fix when the todo is done

PR #123, opened 2023-11-19
improve iOS notification behavior and also remove notifications

ldab:
1. The source comes as `com.facebook.Messenger` so remove the first 2x dots;
2. Use the right UID
3. Allow to remove notifications
4. Had an uninitiated struct that I think it was sending weather data unintentionally.

### DIFF
--- a/app/src/ble/ble_ancs.c
+++ b/app/src/ble/ble_ancs.c
@@ -11,10 +11,12 @@
  *
  * @see https://developer.apple.com/library/archive/documentation/CoreBluetooth/Reference/AppleNotificationCenterServiceSpecification/Specification/Specification.html
  */
+#include "string.h"
 
 #include <zephyr/kernel.h>
 #include <zephyr/logging/log.h>
 #include <zephyr/sys/byteorder.h>
+#include <zephyr/zbus/zbus.h>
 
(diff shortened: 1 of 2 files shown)


app/src/ble/ble_ancs.c
```diff
@@ -459,6 +434,20 @@ static void bt_ancs_notification_source_handler(struct bt_ancs_client *ancs_c,
...
+
+        if (notification_latest.evt_id == BT_ANCS_EVENT_ID_NOTIFICATION_REMOVED) {
+            ble_comm_cb_data_t cb = {
+                .type = BLE_COMM_DATA_TYPE_NOTIFY_REMOVE,
+                .data.notify_remove.id = notification_latest.notif_uid,
+            };
+
+            LOG_DBG("Remove notification %d", cb.data.notify_remove.id);
+
+            data_parsed_cb(&cb);
```
jakkra, line 446: Should use zbus here for not remove maybe reason it doesn't work. See https://github.com/jakkra/ZSWatch/blob/601d09402caafb131a08c64a24114836172fdac6/app/src/ble/ble_comm.c#L747

jakkra (approved):

Issue #112, opened 2023-11-16
Common API for communication back to phone apps


jakkra:
This can remove `#ifdefs` in all application code that sends information back to a phone.

Suggested to create a header file with necessary functions such as "next track" etc.

Then we have two implemementations for it, one for Gadgetbridge and one for iOS.
We choose which one to compile depending on if the compile for Android or iOS

ldab:
I think the FW should be build for both phones and from the connection it can know what type of command to use, no ifdefs, no different headers

jakkra:
> I think the FW should be build for both phones and from the connection it can know what type of command to use, no ifdefs, no different headers

Yes true, can be dynamic, it's easily detectable as we know directly if the phone have the apple services, it's an iPhone.

ldab:
I was thinking we can use the company identifier to determine the device.

alternatively hook the play/pause etc action to a Zbus fire and forget, down the line the GATT client will not be able to write if the service is not available, sanitising it that way, very simple, I do not see the downside of doing it that way

Kampi:
Maybe we can add both variants? Build up Android and iOS separately, enable both by default, and let the user decide to disable one of those to save code space. IMHO for me iOS isn´t relevant and I don´t need it, because I´m using Android.

jakkra:
I don't care which way, the only reason I see to have it a compile thing is that it will not work with upstream Zephyr and need NCS, and that is handy as sometimes when needed to try some patch etc.
But should not make it much more complicated to keep them compile options but allow both, just need to think a bit how to structure it.

ldab:
I haven’t tried but theoretically one can set the zephyr downstream version if needed on the west.yml

jakkra:
You can, but often it won't compile as the ncs stuff is not compatilbe with newer zephyr

PR #126, opened 2023-11-20
Move music control to Zbus and remove `#ifdef`

ldab:
`ble_comm.c` could be refactored and some of it moved to `ble_gadgetbridge.c` for example, `music_control_event_callback` looks a bit misplaced now, but it should work

https://github.com/jakkra/ZSWatch/issues/112 ?

### DIFF
--- a/app/src/applications/music_control/music_control_app.c
+++ b/app/src/applications/music_control/music_control_app.c
@@ -3,13 +3,10 @@
 #include <zsw_clock.h>
 #include <zephyr/zbus/zbus.h>
 
-#ifdef CONFIG_BLE_USES_IOS
-#include "ble/ble_ams.h"
-#endif
-
 #include "music_control_ui.h"
 #include "ble/ble_comm.h"
 #include "events/ble_data_event.h"
+#include "events/music_event.h"
 #include "managers/zsw_app_manager.h"
 
 // Functions needed for all applications
@@ -22,6 +19,8 @@ static void zbus_ble_comm_data_callback(const struct zbus_channel *chan);
 static void handle_update_ui(struct k_work *item);
 
 ZBUS_CHAN_DECLARE(ble_comm_data_chan);
+
+ZBUS_CHAN_DECLARE(music_control_data_chan);
 ZBUS_LISTENER_DEFINE(music_app_ble_comm_lis, zbus_ble_comm_data_callback);
 
 static K_WORK_DEFINE(update_ui_work, handle_update_ui);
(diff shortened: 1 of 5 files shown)


jakkra:
I'm thinking if using zbus for stuff like this is necessary. It's not like this info needs to be propogated around the software, it just goes to either phone "backend". Maybe just a simple API is easier.
For example:
@ldab @Kampi What do you think?
```
// ble_comm.c
#define MAX_BACKENDS 2
ble_comm_backend_cbs backends[MAX_BACKENDS];

ble_comm_register_backend(backend_cbs);

ble_comm_music_control(action) {
    for (int i = 0; i < MAX_BACKENDS; i++) {
        if (ble_comm_backend_cbs[i].music_control) {
            ble_comm_backend_cbs[i].music_control(action);
        }
    }
}

// ble_comm_ios.c

static ble_comm_backend_cbs api = {
    .music_control = music_control,
    .send_battery = send_battery,
    ...
};

ble_comm_register_backend(api);

// ble_comm_gadgetbridge.c

static ble_comm_backend_cbs api = {
    .music_control = music_control,
    .send_battery = send_battery,
    ...
};

ble_comm_register_backend(api);
```

ldab:
> I'm thinking if using zbus for stuff like this is necessary. It's not like this info needs to be propogated around the software, it just goes to either phone "backend". Maybe just a simple API is easier. For example: @ldab @Kampi What do you think?
> 
> ```
> // ble_comm.c
> #define MAX_BACKENDS 2
> ble_comm_backend_cbs backends[MAX_BACKENDS];
> 
> ble_comm_register_backend(backend_cbs);
> 
> ble_comm_music_control(action) {
>     for (int i = 0; i < MAX_BACKENDS; i++) {
>         if (ble_comm_backend_cbs[i].music_control) {
>             ble_comm_backend_cbs[i].music_control(action);
>         }
>     }
> }
> 
> // ble_comm_ios.c
> 
> static ble_comm_backend_cbs api = {
>     .music_control = music_control,
>     .send_battery = send_battery,
>     ...
> };
> 
> ble_comm_register_backend(api);
> 
> // ble_comm_gadgetbridge.c
> 
> static ble_comm_backend_cbs api = {
>     .music_control = music_control,
>     .send_battery = send_battery,
>     ...
> };
> 
> ble_comm_register_backend(api);
> ```

What is the downside of using zbus? I was under the impression that the thing was moving towards not using callbacks, but both solution works.

jakkra:
Possible downside, maybe readability and more boilerplate code to define the zbus events etc.
But on the other hand zbus already handles the code needed in my proposal.

What do you think of just using one event for all phone communication, that will create less boilerplate zbus code. The data type can be an union with an ID?

ldab:
> Possible downside, maybe readability and more boilerplate code to define the zbus events etc. But on the other hand zbus already handles the code needed in my proposal.
> 
> What do you think of just using one event for all phone communication, that will create less boilerplate zbus code. The data type can be an union with an ID? @ldab

I believe @Kampi is working on that on #112

jakkra:
Ok, I'll merge this and then it can be cleaned up to make it same everywhere

Kampi:
@jakkra sounds like a good idea. I had the idea already for BLE connected and disconnected event but it can be expanded with additional stuff. This can also provide some sort of API for the Bluetooth communication in other apps.

jakkra:
> @jakkra sounds like a good idea. I had the idea already for BLE connected and disconnected event but it can be expanded with additional stuff. This can also provide some sort of API for the Bluetooth communication in other apps.

Nice, BUT for the Bluetooth stuff, I think apps should use Zephyr API directly, it has good methods to get callbacks on connect etc. already. This should be used. We can add other stuff related to connections on zbus if we see a need.

PR #129, opened 2023-11-20
Ble enhancements#124

Kampi:
BLE: 

- Add "Subject" field to notification object
- Remove length check for notification fields, because empty fields have a length of 0 which causes issues with the string termination
- Add log level for Bluetooth code
- Replace include guards with "pragma once"

main:
- Rename log level for main in Kconfig
- Add log level for main application
- Use short payload of a notification instead of the long version

Template app:
- Add log level for template app

Settings app:
- Add log level for settings app


app/src/applications/settings/settings_app.c
```diff
@@ -220,6 +231,13 @@ static void on_pairing_enable_changed(lv_setting_value_t value, bool final)
 {
     if (final) {
         ble_comm_set_pairable(true);
+
+        if (!k_work_delayable_is_pending(&ble_pairing_dwork)) {
```
jakkra, line 235: You can replace all of this with one in the top of file:
`K_WORK_DELAYABLE_DEFINE(ble_pairing_dwork, ble_pairing_work_handler)`

And then always call `k_work_reschedule` here, I think it will schedule if not already scheduled, and re-schedule i.e. prolong to 60s

Kampi: Changed

app/src/main.c
```diff
@@ -353,7 +353,7 @@ static void open_notification_popup(void *data)
         lv_group_set_default(temp_group);
         lv_indev_set_group(enc_indev, temp_group);
         zsw_vibration_run_pattern(ZSW_VIBRATION_PATTERN_NOTIFICATION);
-        zsw_notification_popup_show(not->title, not->body, not->src, not->id, on_popup_notifcation_closed, 10);
+        zsw_notification_popup_show(not->title, not->sender, not->src, not->id, on_popup_notifcation_closed, 10);
```
jakkra, line 356: Sender is body? What changes here?

PR #131, opened 2023-11-21
 Remove #ifdef from clock init and also remove "TZ" 

ldab:


app/src/ble/ble_comm.c
```diff
@@ -481,6 +491,33 @@ static void parse_time(char *start_time)
...
+    memset(&cb, 0, sizeof(cb));
+
+    end_timezone = strstr(offset, ")");
+    if (end_timezone) {
+        cb.data.time.tz_offset = strtof(offset, &end_timezone);
+
+        if (offset != end_timezone) {
+            cb.type = BLE_COMM_DATA_TYPE_SET_TIME;
+            LOG_DBG("set time offset: %.1f", cb.data.time.tz_offset);
+            send_ble_data_event(&cb);
```
jakkra, line 516: When you send this event, the time.seconds will be 0, hence causing wrong time. I htink we need a different event type for timezone set or similar.

ldab: Strange, because if I send “GMT2” or other combination it kind of works

jakkra: Let me try again, I see now there is a check if seconds > 0. So not, that, but I got a time of hour 64 hehe.

jakkra: Nope, still getting invalid time. Another thing needed is to put this timezone in retained RAM, as when the watch crashes for example, before gadgetbridge connects again, the time will still be kindof correct. I can fix this if you want, otherwise here is a draft patch for it. Let me know if you want me to do it./home/jakkra/Documents/ZSWatch/diff.txt

jakkra: Fixed it:
GMT+0100 -> NOT OK
GMT+01 -> OK

jakkra: Still gettting wrong time :/ setenv not working as expected TZ not working.
Hardcoding GMT+01 gives two hours back.

ldab: Yeps, that I have tried… se no TZ seems ok tho, at least on posix

jakkra: Ye this is on target, I think newlib-nano is used. I'm trying to find what it wants.

PR #133, opened 2023-11-21
Merge iOS version to Android and have both enabled

ldab:
I didn't enabled at the beginning before I had a chance to try both, tested today, it seems ok:

## POSIX + Android
image 
image

## POSIX + iOS
image

PR #134, opened 2023-11-21
Draft:Ble enhancements#124

Kampi:
- Remove Gadgetbridge code from BLE communication code and split them


Kampi:
Testing draft PR

app/boards/arm/zswatch_nrf5340/zswatch_nrf5340_cpuapp_defconfig
```diff
@@ -27,4 +27,8 @@ CONFIG_PINCTRL=y
 
 CONFIG_CLOCK_CONTROL_NRF_K32SRC_RC=y
 # Ensure GPIO P0.01 and P0.00 can be used
-CONFIG_SOC_ENABLE_LFXO=n
\ No newline at end of file
+CONFIG_SOC_ENABLE_LFXO=n
+
+CONFIG_BT_DIS=y
```
ldab, line 32: This is already defined on `prj.conf` does it need to add again? I don't know what it does

app/src/ble/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 target_sources(app PRIVATE ble_aoa.c)
 target_sources(app PRIVATE ble_comm.c)
 target_sources(app PRIVATE ble_transport.c)
+target_sources(app PRIVATE gadgetbridge/ble_gadgetbridge.c)
 target_sources_ifdef(CONFIG_BLE_USES_IOS app PRIVATE ble_ams.c)
```
ldab, line 5: On #133 I have changed those to add iOS and Android, I will merge locally and test

app/src/ble/ble_ancs.c
```diff
@@ -414,11 +412,16 @@ static int parse_notify(const struct bt_ancs_attr *attr)
 
         // the last message is Negative action label, send only when all data is received;
         case ATTR_ID_NEGATIVE_ACTION_LABEL:
+            struct ble_data_event evt;
```
ldab, line 415: This will not work, the `evt` needs to be static to the `parse_notify()` scope because we build the package in 3/4 different messages

app/src/ble/ble_ancs.c
```diff
@@ -414,11 +412,16 @@ static int parse_notify(const struct bt_ancs_attr *attr)
...
+
             cb.type = BLE_COMM_DATA_TYPE_NOTIFY;
-            data_parsed_cb(&cb);
+
+            memcpy(&evt.data, &cb, sizeof(ble_comm_cb_data_t));
+
+            zbus_chan_pub(&ble_comm_data_chan, &evt, K_MSEC(250));
+
             memset(&cb, 0, sizeof(cb));
```
ldab, line 423: `cb` still remains, needs to remove, no?

app/src/events/ble_event.h
```diff
@@ -5,3 +5,10 @@
 struct ble_data_event {
     ble_comm_cb_data_t data;
 };
+
+struct ble_connect_event {
```
ldab, line 9: I think instead of sending an empty struct you could use `zbus_chan_notify()` instead that send an empty message, it is more elegant

app/src/main.c
```diff
@@ -123,6 +122,8 @@ K_WORK_DEFINE(input_work, run_input_work);
 ZBUS_CHAN_DECLARE(ble_comm_data_chan);
 ZBUS_LISTENER_DEFINE(main_ble_comm_lis, on_zbus_ble_data_callback);
 
+ZBUS_SUBSCRIBER_DEFINE(ble_data_subscriber, 4);
```
ldab, line 125: Question to self, what is the maximum priority? Maybe it should go to some `#define` of some sort

app/src/main.c
```diff
@@ -291,6 +292,23 @@ int main(void)
     // it has the required amount of stack.
     k_work_submit(&init_work);
 
+    while (1) {
+        const struct zbus_channel *chan;
+
+        if (!zbus_sub_wait(&ble_data_subscriber, &chan, K_FOREVER)) {
+            struct ble_data_event evt;
+
+            zbus_chan_read(chan, &evt, K_NO_WAIT);
```
ldab, line 301: this is just for debugging, because it doesn't see to do much

app/boards/arm/zswatch_nrf5340/zswatch_nrf5340_cpuapp_defconfig
Kampi: I remove it from `proj.conf`. The idea is that the configurations for all revisions should be in the default configuration and not in `proj.conf`.

jakkra: Don't follow, prj.conf is the default config?

app/src/ble/ble_ancs.c
Kampi: Changed

app/src/ble/ble_ancs.c
Kampi: I think it can be removed. I did it simple and replaced the callback with `zbus_chan_pub`

app/src/events/ble_event.h
Kampi: Good idea. For now the struct is only a place holder, because I´m not sure what data are relevant to publish after opening a connection.

app/src/main.c
Kampi: It´s not the priority. It´s the length of the queue. 4 is the default value from an example without thinking about it. A define might be a good idea.

https://docs.zephyrproject.org/apidoc/latest/group__zbus__apis.html#gac17a735cccecfc90f26127e48cf6279a

app/src/main.c
Kampi: Yes. Will be removed in the future.

app/src/ble/ble_ancs.c
ldab, line 1: It is not working as it is.

The `cb` on line 401 needs to go and be replaced by the `evt`, I can suggest:
1. Just do the zbus event handler and I fix it later;
2. Give me access to your branch and I can push to it;

ldab:
As a side comment, I'm getting tons of whatsapp messages and I don't even have whatsapp on this phone hahaha so something is off

Kampi:
> As a side comment, I'm getting tons of whatsapp messages and I don't even have whatsapp on this phone hahaha so something is off
> 
> But on the bright side, the "Notification" screen is nice

That´s weird :D
I guess it happens on iOS? Can you please set `CONFIG_ZSW_BLE_LOG_LEVEL_DBG=y` and `CONFIG_NOTIFICATION_APP_LOG_LEVEL_DBG=y` so we can figure out where they come from? WhatsApp isn´t added as a notification source in `zsw_notification_manager`. Only as check in `notification_ui.c`. So maybe some wrong values.

app/src/ble/ble_ancs.c
Kampi: @jakkra can you do it? I don´t have access to Settings, so I can not change it.

app/src/events/ble_event.h
Kampi: I have removed the `connected` and `disconnected` events.

jakkra:
> > As a side comment, I'm getting tons of whatsapp messages and I don't even have whatsapp on this phone hahaha so something is off
> > But on the bright side, the "Notification" screen is nice
> 
> That´s weird :D I guess it happens on iOS? Can you please set `CONFIG_ZSW_BLE_LOG_LEVEL_DBG=y` and `CONFIG_NOTIFICATION_APP_LOG_LEVEL_DBG=y` so we can figure out where they come from? WhatsApp isn´t added as a notification source in `zsw_notification_manager`. Only as check in `notification_ui.c`. So maybe some wrong values.

Many of mine are also WhatsApp hehe and I don't have it

app/src/ble/ble_ancs.c
ldab: Pushed

Kampi:
New update available 👍 

- Improve stability
- Improve the UI
- Improve support for long messages via scrolling (also improves the scrolling through the list, because we don´t have an anchor from the text area)

image

Kampi:
Moved to PR from my fork

PR #141, opened 2023-11-23
Fix display glitching. Randomly turning white or going black.

jakkra:
Issue was that we would send display enter sleep, while we where at the same tie flushing data to it to display. Fixed by making sure rendering finishes and giving the flushing thread some time to flush all data to the display before we till the display to go to sleep.


Kampi:
Whuhu finally :D

jakkra:
> Whuhu finally 👍

It just came to me while walking home today haha :D

PR #144, opened 2023-11-24
Add popup with yes/no and add setting to erase whole external flash

jakkra:
Will fix native_posix compile errors tomorrow.


Kampi:
Can we add both? A popup without yes / no dialog and one with yes / no? So we can use the one without buttons for simple messages like "Here for your information" and the complex one for additional user input. Does it make sense to you?

PR #147, opened 2023-11-25
tools: Point astyle plugin to the astyle.cfg.

jakkra:
@ldab Think this also makes sense to add right? Picks up the correct asyle.cfg automatically now.


ldab:
I thought it was done altomagicaly, but if not, then it is a must

ldab (approved):

PR #148, opened 2023-11-26
Powerpoint remote App

ldab:
1. create a power point remote control app
2. implement HID over BLE


Kampi:
Uh that is cool :D

app/prj.conf
```diff
@@ -135,7 +135,7 @@ CONFIG_BT_DEVICE_APPEARANCE=194
 CONFIG_BT_KEYS_OVERWRITE_OLDEST=y
 CONFIG_BT_SETTINGS=y
 CONFIG_BT_MAX_PAIRED=5
-CONFIG_BT_MAX_CONN=1
+CONFIG_BT_MAX_CONN=2
```
ldab, line 138: @jakkra what do you think? yay or nay?

jakkra: Yay, but there is some code assuming only one connection can happen. So doing a 2nd connection will break some functionality as the new connection will be used. Things like this for example: https://github.com/jakkra/ZSWatch/blob/main/app/src/ble/ble_comm.c#L336

Maybe we need to go over those places and fix that first as otherwise functionality will break when 2nd connection connects.

app/src/ble/ble_hid.c
```diff
@@ -0,0 +1,209 @@
...
+static void key_release_handle(struct k_work *item)
+{
+    int8_t report[8];
+    memset(report, 0, sizeof(report));
+    bt_gatt_notify(NULL, &hog_svc.attrs[5], report, sizeof(report));
+}
+
+static void send_key_release(void)
+{
+    if (k_work_delayable_is_pending(&key_release)) {
```
jakkra, line 169: All those lines can be replaced by k_work_reschedule

app/src/ble/ble_hid.c
```diff
@@ -0,0 +1,209 @@
...
+    // Byte 1: Reserved
+    // Byte 2: Keypress
+    int8_t report[8] = {0x00, 0x00, HID_KEY_RIGHT_ARROW};
+    int ret = bt_gatt_notify(NULL, &hog_svc.attrs[5], report, sizeof(report));
+
+    if (ret != 0) {
+        LOG_ERR("ble_hid_next err: %d", ret);
+    }
+
+    send_key_release();
```
jakkra, line 193: If `bt_gatt_notify` fails, don't send `send_key_release`

app/src/ble/ble_hid.c
```diff
@@ -0,0 +1,209 @@
...
+
+    k_work_schedule(&key_release, K_MSEC(250));
+}
+
+void ble_hid_init(void)
+{
+    ;
+}
+
+void ble_hid_next(void)
```
jakkra, line 181: Propogate the return codes of this "ble hid api", you can ignore the return values when using the functions instead. But makes it a bit cleaner in case the user want to handle errors.

jakkra:
- Works good, and cool animations :) but sometimes the fancy button animation stops in an extended mode.
- If I exit and re-open the app the timer won't restart.
- When display goes to sleep, the `lv_timers` will stop counting, this cause the timer to stop and not increment time when display is off. I suggest to save k_uptime_get when starting, and then every lv_timer call, use `uptime_now - start_uptime` and use that as the time cunter instead.

IMG20231127002850
If you exit the app in this state it also crashes.

ldab:
> Choose what you want to fix now, I suggest at least to fix the timer to count when display is off before we merge this one.

It should be better now, timer is fixed.

PR #161, opened 2023-11-28
Sensor fusion + Speedometer app

ldab:
1. Implement sensor fusion based on https://github.com/xioTechnologies/Fusion
2. A crude speedometer app (haven't really tested on motion) it is cold outside :cold_face: 
3. few todo's but I think we can keep going?

PS. I wanted to keep the screen on, but It does not work, do you know why?

image


Kampi:
@ldab the screen is handled by `zsw_power_manager`. We don´t have an option to keep the display on yet.

https://github.com/jakkra/ZSWatch/blob/879fbebf9c3e1d4c1cb41e9e2a49df81db8737fc/app/src/managers/zsw_power_manager.c#L61

jakkra:
> @ldab the screen is handled by `zsw_power_manager`. We don´t have an option to keep the display on yet.
> 
> https://github.com/jakkra/ZSWatch/blob/879fbebf9c3e1d4c1cb41e9e2a49df81db8737fc/app/src/managers/zsw_power_manager.c#L61

We do, I have added an option in settings for it, but somehow it's broken now.

Edit: You need to restart for it to take effect.

PS. I think the new flick out need to check for this setting

Kampi:
What unit is the speedometer using? :D

Kampi:
> > What unit is the speedometer using? 😄
> 
> km/h

Can you add a label with the unit to the app?

PR #163, opened 2023-11-28
Flick In event don´t put the device into inactive mode when the display always on is enabled

Kampi:


ldab:
It works for me, when looking at you did, I was expecting to get the settings by example `settings_storage_get()` but whatever you did, works haha

jakkra:
@ldab this PR should be OK, at least from looking at it. It's a different feature if apps want to keep the screen from going to sleep. That would be implemented some other way. It should not control it though settings, this should be done in another way

Kampi:
> ~It works for me~, when looking at you did, I was expecting to get the settings by example `settings_storage_get()` but whatever you did, ~works haha~.
> 
> Hold on... doesn't work here, I think the debugger was keeping the screen on

Are you sure? I tested it again and the screen on my watch is active for ~2h now, even with flick events.

Issue #124, opened 2023-11-19
BLE enhancements


Kampi:
Collection of different BLE topics

**ToDo**

- [x] Add timeout for enabled pairing
- [x] Fix bug with wrong JSON splitting in `ble_comm.c`
- [x] #48
- [ ] #54
- [x] #109  
- [ ] #112 
- [ ] #72
- [x] Update device information for Device Information service
- [x] Can we move all Gadgetbridge related stuff in a separate file?
- [x] Do we really need to connect and disconnect on ZBUS because apps should use Zephyr API (decide when we close the PR)?

**Open issues with notifications**

- [x] Crashing when leaving the notifications list when a notification is available
- [x] Longer notifications the body can't be seen (I guess LVGL scrolls to the bottom or something)
- [x] Sometimes the age of notification shows a huge number
- [x] Easy to accidentally remove the notification when scrolling it
- [x] Convert the age into a nice time format
- [x] Scrolling of the body field is strange
- [x] Sometimes the body field will be focused
- [x] Scrolling the notification list can be a little bit weird. I guess because of the scrolling body field. Need to check it in detail
- [ ] Hide notification popup when the app is open. Here we should think about an API to get the app status
- [ ] Notification is not 100% centered
- [ ] [Scroll one](https://docs.lvgl.io/master/overview/scroll.html#scroll-one)
- [x] Notification removal on the phone doesn´t work
- [ ] Support for YouTube notifications
- [x] Support for WhatsApp notificaitons

Kampi:
@ldab 

I split the Gadgetbridge related code up into a separate source file.

https://github.com/jakkra/ZSWatch/tree/BLE_enhancements%23124

ldab:
Will check the PR tomorrow 👊

Kampi:
We are changing the communication flow for notifications to this
![image](https://github.com/jakkra/ZSWatch/assets/3458587/7c12a43e-8114-4133-97a5-0b3d574ed439)

This allows us to create and push internal notifications to the notification manager. The notification manager is the main instance in handling all notifications.

@jakkra @ldab 
I pushed the latest version of the notification rework. Feel free to test this version and give some feedback :)
Not 100% nice now, but it´s working.

**Changes so far**

- New layout
- Different icons for WhatsApp, Mail, etc.
- Title and body field
- Scrollable body (not sure if I will keep it)
- Age of the notification
- Scrolling improvement
- Notifications will be added to the notification app when the app is open

**Known changes**

[...]

PR #166, opened 2023-11-29
Trivia app to demonstrate HTTP requests

ldab:
Easy trivia app getting questions from https://opentdb.com/api_config.php

1. It demonstrates the HTTP request thru Gadgetbridge
2. Adds a cool app

Todo:
- [ ] ~~It only works once for some reason~~it is just slow, may take more questions in one request
- [ ] Merge the ble changes to #124 
- [ ] Prompt when iOS user tries it
- [ ] Handle errors
- [ ] Handle html format for example `&#039;`


jakkra:
Haha this is so cool! Awesome work!
I think parsing HTTP responses is a reason to turn on Zephyr JSON module to be able to build some more advanced stuff using the HTTP api in GB: https://docs.zephyrproject.org/apidoc/latest/group__json.html

Kampi:
> Haha this is so cool! Awesome work! I think parsing HTTP responses is a reason to turn on Zephyr JSON module to be able to build some more advanced stuff using the HTTP api in GB: https://docs.zephyrproject.org/apidoc/latest/group__json.html
> 
> Not needed now, but it's something to add together with a generic HTTP API that maps to GB

Then we also could replace the self-made JSON parsing for Gadgetbridge :)

jakkra:
> we also could replace the self-made JSON pa

Yep if we want to since it will be there anyway.

PR #172, opened 2023-11-30
Ble enhancements#124

Kampi:


app/src/applications/notification/ui_export/screens/notification_ui.c
```diff
@@ -0,0 +1,301 @@
...
+    lv_obj_clear_flag(ui_LabelTimeDelta,
+                      LV_OBJ_FLAG_PRESS_LOCK | LV_OBJ_FLAG_CLICK_FOCUSABLE | LV_OBJ_FLAG_GESTURE_BUBBLE | LV_OBJ_FLAG_SNAPPABLE |
+                      LV_OBJ_FLAG_SCROLLABLE | LV_OBJ_FLAG_SCROLL_ELASTIC | LV_OBJ_FLAG_SCROLL_MOMENTUM | LV_OBJ_FLAG_SCROLL_CHAIN);
+    lv_obj_set_style_text_color(ui_LabelTimeDelta, lv_color_hex(0x8C8C8C), LV_PART_MAIN | LV_STATE_DEFAULT);
+    lv_obj_set_style_text_opa(ui_LabelTimeDelta, 255, LV_PART_MAIN | LV_STATE_DEFAULT);
+
+    active_notifications[active_notification_num].panel = ui_Panel;
+    active_notifications[active_notification_num].deltaLabel = ui_LabelTimeDelta;
+    active_notifications[active_notification_num].notification = not;
+    active_notifications[active_notification_num].timer = lv_timer_create(label_on_Timer_Callback, 1000UL,
```
jakkra, line 155: May be a bit unescessary of having one timer per notification, just having one that runs every second and it just loops through all notifications and updated their time would  be a bit more efficient. Could also set it to every 5s to save some CPU load.
Update if you want. Fine anyway.

app/src/ble/ble_comm.c
```diff
@@ -458,473 +410,4 @@ static void bt_receive_cb(struct bt_conn *conn, const uint8_t *const data, uint1
         LOG_DBG("%s", receive_buf);
         parse_data(receive_buf, parsed_data_index);
     }
-}
```
jakkra, line 461: I'd take all the content of `bt_receive_cb` and move it into ble_gadgetbridge.c in a function `ble_gadgetbridge_input(data, len)`

app/src/main.c
```diff
@@ -119,6 +119,9 @@ K_WORK_DEFINE(input_work, run_input_work);
 
 ZBUS_CHAN_DECLARE(ble_comm_data_chan);
 ZBUS_LISTENER_DEFINE(main_ble_comm_lis, on_zbus_ble_data_callback);
+ZBUS_LISTENER_DEFINE(main_notification_lis, on_zbus_notification_callback);
+
+ZBUS_SUBSCRIBER_DEFINE(ble_data_subscriber, 4);
```
jakkra, line 124: `ble_data_subscriber ` not used anywhere? Can be removed?

app/src/managers/zsw_notification_manager.c
```diff
@@ -1,71 +1,135 @@
...
+}
+
+static void notification_mgr_zbus_ble_comm_data_callback(const struct zbus_channel *chan)
+{
+    // Need to context switch to not get stack overflow.
+    // We are here in host bluetooth thread.
+    const struct ble_data_event *event = zbus_chan_const_msg(chan);
+
+    if (event->data.type == BLE_COMM_DATA_TYPE_NOTIFY) {
+        memcpy(&notification, &event->data.data.notify, sizeof(ble_comm_notify_t));
```
jakkra, line 61: How about calling `zsw_notification_manager_add` here and remvove the `static ble_comm_notify_t notification;`
Then in the k_work instead just send the zbus event?

app/src/managers/zsw_notification_manager.c
```diff
@@ -1,71 +1,135 @@
...
+
+        notifications[idx].src = NOTIFICATION_SRC_WHATSAPP;
         notifications[idx].id = not->id;
-        memcpy(notifications[idx].title, not->title, MIN(not->title_len, NOTIFICATION_MGR_MAX_FIELD_LEN - 1));
-        memcpy(notifications[idx].body, not->body, MIN(not->body_len, NOTIFICATION_MGR_MAX_FIELD_LEN - 1));
-        memcpy(notifications[idx].sender, not->sender, MIN(not->sender_len, NOTIFICATION_MGR_MAX_FIELD_LEN - 1));
+        memcpy(notifications[idx].sender, not->title, MIN(not->title_len, ZSW_NOTIFICATION_MGR_MAX_FIELD_LEN - 1));
+        memcpy(notifications[idx].body, not->body, MIN(not->body_len, ZSW_NOTIFICATION_MGR_MAX_FIELD_LEN - 1));
+    } else if (strncmp(not->src, "YouTube", not->src_len) == 0) {
+
```
jakkra, line 109: Remove this `if` and let youtube end up in the final `else` (or fill in the fields) as this code will now work properly if you happen to get a YouTube notification

app/src/managers/zsw_notification_manager.c
```diff
@@ -74,25 +138,52 @@ zsw_not_mngr_notification_t *zsw_notification_manager_add(ble_comm_notify_t *not
...
+}
+
+void zsw_notification_manager_remove_all(void)
+{
+    for (uint32_t i = 0; i < num_notifications; i++) {
+
+        notifications[i].id = NOTIFICATION_INVALID_ID;
+    }
+
+    zbus_chan_notify(&zsw_notification_mgr_remove_chan, K_NO_WAIT);
```
jakkra, line 170: Is this part still TODO? This would send an unpopulated event, not sure what the listeners would do with this info.
I'd do this in the loop above instead. All listeners will correctly get the callback for each notify and they can all be handled.

jakkra (commented):
Looks good. Fix what you think makes sense.

app/src/applications/notification/ui_export/screens/notification_ui.c
Kampi: Good point. Haven't thought about it.

app/src/main.c
Kampi: Have to check it.

app/src/managers/zsw_notification_manager.c
Kampi: Good point. There is no rule for the listeners to check for an invalid notification. I will change it.

Kampi: Wait - this is a notification-removed event. I defined it in the following way:

```
/** @brief  This event informs all listeners about a removed notification. A notify is
 *          send, when all notifications were removed. Otherwise, the event will contain a copy
 *          of the removed notification.
*/
```

https://github.com/Kampi/ZSWatch/blob/67abc93f2048c2d93218b6126103227a03647da6/app/src/events/zsw_notification_event.c#L13

jakkra: Right ok, but how will ble_comm know which IDs to send to phone that they are marked as read?

Kampi: Removed :)

PR #186, opened 2023-12-09
do not refresh watchface if only one is selected

apewo:
this one line fix prevent the watchface app to refresh/glitch when only one watchface is selected

### DIFF
--- a/app/src/applications/watchface/watchface_app.c
+++ b/app/src/applications/watchface/watchface_app.c
@@ -155,6 +155,10 @@ void watchface_app_stop(void)
 
 void watchface_change(void)
 {
+    if (num_watchfaces == 1) {
+        return;
+    }
+
     watchfaces[current_watchface]->remove();
     current_watchface = (current_watchface + 1) % num_watchfaces;
 


Kampi:
@apewo did you run `west format`? If not please run it and push the change again 👍

Kampi:
@apewo we still have a format issue in `battery_app.c`. I think this is a common issue with newer versions of `astyle` that also happens to me. Which version of `astyle` do you use? If it´s not 3.1 can you please download and use 3.1?

apewo:
the pip, or the native app ?, the native is 3.4 on my distro, i will have to recompile 3.1 then.
the pip , minimum i see is 3.3
```
pip install astyle==3.1
ERROR: Could not find a version that satisfies the requirement astyle==3.1 (from versions: 3.3b3, 3.3, 3.3.1, 3.4, 3.4.2, 3.4.4, 3.4.5, 3.4.6, 3.4.7, 3.4.8, 3.4.9, 3.4.10)
ERROR: No matching distribution found for astyle==3.1
```

jakkra:
> the pip, or the native app ?, the native is 3.4 on my distro, i will have to recompile 3.1 then. the pip , minimum i see is 3.3
> 
> ```
> pip install astyle==3.1
> ERROR: Could not find a version that satisfies the requirement astyle==3.1 (from versions: 3.3b3, 3.3, 3.3.1, 3.4, 3.4.2, 3.4.4, 3.4.5, 3.4.6, 3.4.7, 3.4.8, 3.4.9, 3.4.10)
> ERROR: No matching distribution found for astyle==3.1
> ```

On Ubuntu sudo apt-get only have 3.1, this is why it's the one we use for now. Try install with apt

PR #188, opened 2023-12-09
Reset button adjustments

Kampi:
- Increase long press time for system reset to 10 s to match with nPM1300
- Move system reset button from top left to bottom right


jakkra:
Can you do the same update in the clockwise_rotate.overlay

PR #192, opened 2023-12-09
Add reset for step counter + smaller cleanups

Kampi:
BMI270:
- Add reset for step counter via offset attribute

IMU:
- Add a function to reset the step counter
- Add periodic reset at 00:00 for step counter (needs rework later when RTC is added)

Battery:
- Move "vbat" device tree entry into rev 1-4 files

IAQ app:
- Align app source name with other apps
- Add initial measurement during app startup

Template:
- Fix broken macro in CMakeLists

Misc:
- Remove "sw-top-right", "sw-top-left", "sw-bottom-right" and "sw-bottom-left" because they are unused

### DIFF
--- a/app/boards/arm/zswatch_nrf5340/zswatch_nrf5340_common.dts
+++ b/app/boards/arm/zswatch_nrf5340/zswatch_nrf5340_common.dts
@@ -18,14 +18,6 @@
         zephyr,display = &gc9a01;
     };
 
-    vbatt {
-        compatible = "voltage-divider";
-        io-channels = <&adc 7>;
-        output-ohms = <220000>;
-        full-ohms = <(1500000 + 220000)>;
-        power-gpios = <&gpio1 10 0>;
-    };
-
     pwmleds {
         compatible = "pwm-leds";
         display_blk: pwm_led_0 {
@@ -43,6 +35,11 @@
         invert-x;
     };
 
+    rtc: rtc {
+        status = "okay";
+        compatible = "zephyr,rtc-emul";
+    };
+
     aliases {
         display-blk = &display_blk;
         vibrator-pwm = &vibrator_pwm;
@@ -50,10 +47,6 @@
         sw2 = &button2;
         sw3 = &button3;
         sw4 = &button4;
-        sw-top-right = &button1;
-        sw-top-left = &button4;
-        sw-bottom-right = &button3;
-        sw-bottom-left = &button2;
         watchdog0 = &wdt0;
         mcuboot-button1 = &button1;
     };
@@ -120,7 +113,6 @@
         status = "okay";
         int-gpios = <&gpio0 25 (GPIO_PULL_UP | GPIO_ACTIVE_LOW)>;
     };
-
 };
 
 &i2c2 {
@@ -201,7 +193,6 @@ zephyr_udc0: &usbd {
 };
 
 / {
-
     reserved-memory {
         #address-cells = <1>;
         #size-cells = <1>;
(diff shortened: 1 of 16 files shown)


app/boards/arm/zswatch_nrf5340/zswatch_nrf5340_common.dts
```diff
@@ -18,14 +18,6 @@
         zephyr,display = &gc9a01;
     };
 
-    vbatt {
-        compatible = "voltage-divider";
```
jakkra, line 22: This should be in the common one as it's the same in all or?

app/drivers/sensor/bmi270/bosch_bmi270.c
```diff
@@ -280,13 +291,9 @@ static int bmi270_attr_set(const struct device *p_dev, enum sensor_channel chann
         switch (attribute) {
             case SENSOR_ATTR_CONFIGURATION:
                 if ((p_value->val2 & 0x01) == 0) {
-                    if (bmi2_disable_feature(p_dev, p_value->val1 & 0xFF) != BMI2_OK) {
-                        return -EFAULT;
```
jakkra, line 284: Return code remap should be kept as we go from bosch driver error codes to Zephyr error codes?

app/src/sensors/zsw_imu.c
```diff
@@ -36,8 +38,16 @@ static struct sensor_trigger bmi270_trigger;
 
 static void zbus_periodic_slow_callback(const struct zbus_channel *chan)
 {
+    struct tm *timeinfo;
     struct accel_event evt = {
     };
+
+    timeinfo = zsw_clock_get_time();
+
+    if ((timeinfo->tm_hour == 0) && (timeinfo->tm_min == 0)) {
```
jakkra, line 47: Remove, was for debugging?

app/boards/arm/zswatch_nrf5340/zswatch_nrf5340_common.dts
Kampi: I moved it because rev 5 don't need it anymore.

app/src/sensors/zsw_imu.c
Kampi: No it's the current way to reset the step counter at 00:00. Not the best one but we don't have any alarm function yet. Maybe we implement RTC first and change it to the better solution?

app/drivers/sensor/bmi270/bosch_bmi270.c
Kampi: `bmi2_disable_feature` returns a Zephyr error code already. The if statement was wrong before. Now I return the error code from `bmi2_disable_feature` directly.

app/boards/arm/zswatch_nrf5340/zswatch_nrf5340_common.dts
jakkra: Understood

app/src/sensors/zsw_imu.c
jakkra: Was referring to it not really resting the steps, it only does a print. Should call the reset step function here then. I'd suggest change to 23:59 also, so I this code won't trigger immediately after a reset when flashing new code.

Kampi: Good point. I will change it

Kampi: Oh, now I see it. I forgot the step counter reset. Thought it was there :D

PR #201, opened 2023-12-11
Power consumption improvements

jakkra:
From constant 2mA when idle to this
image

**TODO**

- [x] After `NO_MOTION` (stationary state) and a move back to wearing state, idle current is 10mA instead of 2mA.
- [x] Maybe rename slow, mid and fast perdiodic intervals to `INTERVAL_100_MS, INTERVAL_1000_MS, INTERVAL_10000_MS` or similar names instead.


app/drivers/sensor/bmp581/bosch_bmp581.c
Kampi, line 1: IMO we should use a delayable work here (like in the light sensor) where the delay time is based on the data rate settings. The sensor can measure with a frequency of 0.125 Hz in the slowest settings and this might block.

jakkra: Ok, but the current code did not do that anyway and https://github.com/boschsensortec/BMP5-Sensor-API/blob/c779b44bf3c682d1fc06b5771378361650028223/bmp5.c#L612 this function is not blocking.
But I'll fix it if it's needed

Kampi: Okay so we don´t have any blocking issue here. But we need at least some sort of mechanism to do a reading when a sensor value is ready (like in the light sensor driver). This is a lack of the current BMP581 driver, that´s true. It works currently, because the conversion and readout times are low, but when you reduce the data rate to 0.125 Hz for example then you read out garbage until the conversion is ready and the result valid.

So the best way would be (IMO):
When triggering a new measurement:
- Get ODR from the IC or from local data structure
- Calculate a delay time for the readout
- Schedule a delayable work
- In delayable work read the result and update the latest sensor value with it

So you can read out the latest sensor value between each measurement.

jakkra: Yup will take a look at it

Issue #206, opened 2023-12-21
sdl native posix upscaling


elmo2k3:
Hi,

in your talk you mentioned that you modified display_sdl.c to upscale the sdl output. I tried to place SDL_RenderSetLogicalSize(*renderer,600,600) to different places also in display_sdl_bottom.c but without success. Could you give me a hint?

Cheers
Bjoern

jakkra:
I removed my patch a while ago as stuff in Zephyr was moved around so it wouldn't apply. But in latest Zephyr on main it's actually implemented now. But ncs is not using that version yet. So you can probably find the commit upstream and cherry pick it.

elmo2k3:
Found it, thanks!! Nice and simple display_zoom_pct option.

jakkra:
Nice! If it applies to ZSWatch Zephyr version, feel free to add it here and do a PR https://github.com/jakkra/ZSWatch/tree/main/app/patches

jakkra:
Merged and fixed

PR #207, opened 2023-12-21
Patch to include CONFIG_SDL_DISPLAY_ZOOM_PCT from Zephyr upstream

elmo2k3:
As mentioned in #206


jakkra:
Thanks, I tried it and although the zoom works, I think probably one commit fixing the touch is needed also. The coordinates need to be adjusted according to the zoom. I assume this was also done in combination with this upstream commit, but not sure.

jakkra:
Actually I tried latest Zephyr and it's not working there either. I have submitted a PR to Zephyr fixing this. I'll add that patch to this PR also when it's merged.

jakkra:
This patch works now after upgrading to NCS 2.6, merging.
Will need an additional patch to fix touch coordinates I fixed in Zephyr, coming up in a new PR.

PR #209, opened 2023-12-26
New settings page. Revert to gestures using LVGL

jakkra:
See commits.

### DIFF
--- a/app/src/applications/settings/settings_app.c
+++ b/app/src/applications/settings/settings_app.c
@@ -31,8 +31,11 @@ static void on_pairing_enable_changed(lv_setting_value_t value, bool final);
 static void on_reset_steps_changed(lv_setting_value_t value, bool final);
 static void on_clear_bonded_changed(lv_setting_value_t value, bool final);
 static void on_clear_storage_changed(lv_setting_value_t value, bool final);
+static void on_reboot_changed(lv_setting_value_t value, bool final);
+static void on_restart_screen_changed(lv_setting_value_t value, bool final);
 
 static void ble_pairing_work_handler(struct k_work *work);
+static void display_restart_work_handler(struct k_work *work);
 
 LV_IMG_DECLARE(settings);
 
@@ -45,6 +48,8 @@ typedef struct setting_app {
 } setting_app_t;
 
 K_WORK_DELAYABLE_DEFINE(ble_pairing_dwork, ble_pairing_work_handler);
+K_WORK_DEFINE(display_restart_work, display_restart_work_handler);
+
 
 // Default values.
 static setting_app_t settings_app = {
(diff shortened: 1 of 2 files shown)


Kampi:
I think it´s a good idea to add the following changes:

- Rename "General" to "Display"
- Move "Screen" (what is it exactly?) from "Other" to "Display"
- Move "Vibrate" and "Reset Step Counter" from "Display" to "Other"

Also the "Back" button (top left) doesn´t work. Instead, it scrolls through the menu and swipes up to open "Settings" and also opens "General" often.

jakkra:
> I think it´s a good idea to add the following changes:
> 
> * Rename "General" to "Display"
> * Move "Screen" (what is it exactly?) from "Other" to "Display"
> * Move "Vibrate" and "Reset Step Counter" from "Display" to "Other"
> 
> Also the "Back" button (top left) doesn´t work. Instead, it scrolls through the menu and swipes up to open "Settings" and also opens "General" often.

Good idea, will do like that. 
Back button never worked with touch, but handy for navigating in posix and with buttons. It should be moved a bit.

Kampi:
> > I think it´s a good idea to add the following changes:
> > 
> > * Rename "General" to "Display"
> > * Move "Screen" (what is it exactly?) from "Other" to "Display"
> > * Move "Vibrate" and "Reset Step Counter" from "Display" to "Other"
> > 
> > Also the "Back" button (top left) doesn´t work. Instead, it scrolls through the menu and swipes up to open "Settings" and also opens "General" often.
> 
> Good idea, will do like that. Back button never worked with touch, but handy for navigating in posix and with buttons. It should be moved a bit.

Haven´t tested it with touch. It was the regular button. :P

jakkra:
> > > I think it´s a good idea to add the following changes:
> > > 
> > > * Rename "General" to "Display"
> > > * Move "Screen" (what is it exactly?) from "Other" to "Display"
> > > * Move "Vibrate" and "Reset Step Counter" from "Display" to "Other"
> > > 
> > > Also the "Back" button (top left) doesn´t work. Instead, it scrolls through the menu and swipes up to open "Settings" and also opens "General" often.
> > 
> > 
> > Good idea, will do like that. Back button never worked with touch, but handy for navigating in posix and with buttons. It should be moved a bit.
> 
> Haven´t tested it with touch. It was the regular button. :P

Hmm, the back button in settings work as expected for me.

Kampi:
> > > > I think it´s a good idea to add the following changes:
> > > > 
> > > > * Rename "General" to "Display"
> > > > * Move "Screen" (what is it exactly?) from "Other" to "Display"
> > > > * Move "Vibrate" and "Reset Step Counter" from "Display" to "Other"
> > > > 
> > > > Also the "Back" button (top left) doesn´t work. Instead, it scrolls through the menu and swipes up to open "Settings" and also opens "General" often.
> > > 
> > > 
> > > Good idea, will do like that. Back button never worked with touch, but handy for navigating in posix and with buttons. It should be moved a bit.
> > 
> > 
> > Haven´t tested it with touch. It was the regular button. :P
> 
> Hmm, the back button in settings work as expected for me.

I tested it on rev 3. Let me repeat it and make a video to show it.

PR #216, opened 2024-01-17
Bug fix wrong notification time

Kampi:


Kampi:
mmh @jakkra  what is the reason for these 12 commits? They are old?

jakkra:
Maybe old history on your branch. Did you rebase on main branch? Otherwise checkout main and cherry pick the actual fix commit.

Kampi:
> Maybe old history on your branch. Did you rebase on main branch? Otherwise checkout main and cherry pick the actual fix commit.

I reset my main branch to upstream/main and now it looks good.

PR #239, opened 2024-03-29
about app: Implement the about app.

jakkra:
All filled in dynamically.
image


Kampi:
Where does the version string come from? Do you use it from sdkconfig?

jakkra:
> Where does the version string come from? Do you use it from sdkconfig?

For hw revision comes from the board define. Rest from build system.

PR #246, opened 2024-04-14
Update Kconfig

Kampi:
Set default PCB revision in Kconfig to 5


Kampi:
@jakkra PCB version 5 as default fails in CI/CD. Can we fix it? If I remember correctly you mentioned that the Flash memory causes some issues.

jakkra:
On it

jakkra:
Ok, so you need to set it to verson 3 in native_sim (or keep same default as now) due to not having filesystem support in `native_sim` yet. (Or just skip changing the default and close the PR)

jakkra (approved):

Issue #249, opened 2024-04-15
Implement crash log dumping


Kampi:
- Save dump in RAM for reboot
- Write the dump into Flash after reboot (up to 5 in a circular buffer)
- Add an option to enable download mode in UI
- Download the dump via RTT or USB (Revision 5 or later)

PR #256, opened 2024-04-16
Add coredump support

jakkra:
See `DEBUGGING_ISSUES.md` for how to use it.

Implements parts of https://github.com/jakkra/ZSWatch/issues/249

- Store coredump in RAM area when crashing
- At bootup, check if there is a coredump in RAM
- If so write it to external flash
- And a bunch of other nice things around it.

More coredump improvements to come


app/src/zsw_coredump.c
```diff
@@ -0,0 +1,329 @@
...
+    const char *fileName;
+    struct crash_info_header header;
+
+    fileName = file;
+
+    if (strlen(file) > MAX_FILENAME_LEN) {
+        // Copy end of filename as it's more relevant than the path to the file.
+        fileName += strlen(file) - MAX_FILENAME_LEN;
+    }
+    strncpy(header.crash_file, fileName, MAX_FILENAME_LEN);
```
Kampi, line 294:
```
../src/zsw_coredump.c: In function 'assert_post_action':
../src/zsw_coredump.c:294:5: warning: '__builtin_strncpy' specified bound 32 equals destination size [-Wstringop-truncation]
  294 |     strncpy(header.crash_file, fileName, MAX_FILENAME_LEN);
      |     ^~~~~~~
```

.vscode/settings.json
```diff
@@ -5,7 +5,9 @@
     ],
     "nrf-connect.boardRoots": [
         "${workspaceFolder}/app",
-        "${workspaceFolder}\\app"
+        "${workspaceFolder}\\app",
+        "${workspaceFolder}/zephyr/boards",
+        "${workspaceFolder}\\zephyr\\boards",
```
Kampi, line 10: Why double?

app/src/zsw_coredump.c
jakkra: Why do I not get this warning?...

jakkra: Oh, I get it in release build, but not debug build Need to figure out why. Should show in debug also.

PR #263, opened 2024-04-22
Generic history for different sensors

Kampi:
First idea and not tested.


jakkra (approved):
Looks awesome!

Kampi:
> Looks awesome!

But doesn´t work yet :D

app/src/history/zsw_history.c
```diff
@@ -0,0 +1,186 @@
...
+
+    sprintf(key_data, "%s/%s", p_history->key, ZSW_HISTORY_DATA_EXTENSION);
+    error = settings_load_subtree_direct(key_data, zsw_history_load_data_cb, p_history);
+    LOG_DBG("Load data with key %s", key_data);
+    if (error) {
+        LOG_ERR("Error during data loading! Error: %i", error);
+        return -EFAULT;
+    }
+
+    error = settings_delete(key_header);
```
jakkra, line 149: Don't understand why there is a delete here, do you know why?
Edit: I see it was in the battery app before, but I think that's wrong?

app/src/history/zsw_history.c
```diff
@@ -0,0 +1,186 @@
...
+
+void zsw_history_del(zsw_history_t *p_history)
+{
+    __ASSERT(p_history == NULL, "Invalid parameter for zsw_history_del");
+
+    memset(p_history->samples, 0, p_history->num * p_history->sample_size);
+    p_history->write_index = 0;
+}
+
+void zsw_history_add(zsw_history_t *p_history, const void *p_sample)
```
jakkra, line 92: Should we make this static instead? It's right now only used inside this file.

jakkra: I suggest renaming `zsw_history_save` to `zsw_history_add`, as that's the one actually doing the adding and saving of a new sample. Think it makes it a bit more clear when used.

app/src/history/zsw_history.c
```diff
@@ -0,0 +1,186 @@
...
+    error = settings_subsys_init();
+    if (error) {
+        LOG_ERR("Error during settings initialization! Error: %i", error);
+        return -EFAULT;
+    }
+
+    return 0;
+}
+
+void zsw_history_del(zsw_history_t *p_history)
```
jakkra, line 84: Should `zsw_history_del` also delete the settings? Otherwise let's say you call `zsw_history_del` and then reset the watch before a new sample was added, then next time watch boots, it will read in the old samples.

app/src/history/zsw_history.c
```diff
@@ -0,0 +1,186 @@
...
+
+int zsw_history_load(zsw_history_t *p_history)
+{
+    int32_t error;
+
+    __ASSERT((p_history != NULL) &&
+             (strlen(p_history->key) <= ZSW_HISTORY_MAX_KEY_LENGTH), "Invalid parameters for zsw_history_load");
+
+    sprintf(key_header, "%s/%s", p_history->key, ZSW_HISTORY_HEADER_EXTENSION);
+    error = settings_load_subtree_direct(key_header, zsw_history_load_header_cb, p_history);
```
jakkra, line 131: For me `zsw_history_load_header_cb` never gets called, and even with an erased flash it says max samples stored.

app/src/history/zsw_history.c
```diff
@@ -0,0 +1,186 @@
...
+    LOG_DBG("   Num: %u", p_history->num);
+    LOG_DBG("   Sample size: %u", p_history->sample_size);
+    LOG_DBG("   Write index: %u", p_history->write_index);
+    if (error) {
+        LOG_ERR("Error during header loading! Error: %i", error);
+        return -EFAULT;
+    }
+
+    sprintf(key_data, "%s/%s", p_history->key, ZSW_HISTORY_DATA_EXTENSION);
+    error = settings_load_subtree_direct(key_data, zsw_history_load_data_cb, p_history);
```
jakkra, line 142: Same, `zsw_history_load_data_cb` not called for me

jakkra (changes requested):

app/src/applications/battery/battery_app.c
```diff
@@ -110,38 +95,12 @@ static void zbus_battery_sample_data_callback(const struct zbus_channel *chan)
 
 static void on_battery_hist_clear_cb(void)
 {
-    memset(battery_context.samples, 0, sizeof(battery_context.samples));
+    zsw_history_del(&battery_context);
     if (settings_delete(SETTING_BATTERY_HIST) != 0) {
```
jakkra, line 99: Move this delete into `zsw_history_del`

app/src/history/zsw_history.c
Kampi: It depends if you want to keep the history after loading it.

app/src/history/zsw_history.c
Kampi: The idea was to use `zsw_history_add` for adding a sample to a history without saving the history on the file system. `zsw_history_save` can be used for storing it also on the file system.

app/src/history/zsw_history.c
jakkra: Think you always want. Don't see why not. Also you may run into issues when you load and then reset the watch before a save has been done. Then all data will be lost.

app/src/history/zsw_history.c
Kampi: Good point.

app/src/history/zsw_history.c
jakkra: My bad, working

app/src/history/zsw_history.c
jakkra: My bad, working

app/src/history/zsw_history.c
jakkra: Ok, I see, good. Could update save to not take any arg and require the user to first add and then call save?
I see uses for add and the one save when done. For example when sampling alot of data, you sample and when done you call save. OR maybe allow input NULL to `zsw_history_save` to just make a save without adding, maybe the most flexible way!

jakkra:
I think we should include num_stored_samples in the `zsw_history_t` to allow the user to know how many of the samples are actually valid. Since the `num` param today is always the max number of samples. Or can `num` be changed to actually reflect the number of valid samples?

jakkra:
> I think we should include num_stored_samples in the `zsw_history_t` to allow the user to know how many of the samples are actually valid. Since the `num` param today is always the max number of samples. Or can `num` be changed to actually reflect the number of valid samples?

Because in the battery app now it plots all samples, which are all 0 so it's hard to see anything.

app/src/history/zsw_history.c
Kampi: Okay so I remove it.

app/src/history/zsw_history.c
Kampi: Or maybe split "add" and "save"? So you use "add" for adding values to the history and then you can call save to save the history but save doesn´t add a value to the history. I think this is the more intuitive way?

app/src/history/zsw_history.c
Kampi: You can call `zsw_history_del` after loading it if you don´t want to keep it.

Kampi:
> > I think we should include num_stored_samples in the `zsw_history_t` to allow the user to know how many of the samples are actually valid. Since the `num` param today is always the max number of samples. Or can `num` be changed to actually reflect the number of valid samples?
> 
> Because in the battery app now it plots all samples, which are all 0 so it's hard to see anything.

I think we should keep `num` as an indicator for the maximum size of the buffer. But sure we can add a counter for the added samples.

Kampi:
> > I think we should include num_stored_samples in the `zsw_history_t` to allow the user to know how many of the samples are actually valid. Since the `num` param today is always the max number of samples. Or can `num` be changed to actually reflect the number of valid samples?
> 
> Because in the battery app now it plots all samples, which are all 0 so it's hard to see anything.

I think we should keep `num` (but rename it to length) to have an indicator about the maximum length. Then we add a sample counter.

jakkra:
> > > I think we should include num_stored_samples in the `zsw_history_t` to allow the user to know how many of the samples are actually valid. Since the `num` param today is always the max number of samples. Or can `num` be changed to actually reflect the number of valid samples?
> > 
> > 
> > Because in the battery app now it plots all samples, which are all 0 so it's hard to see anything.
> 
> I think we should keep `num` (but rename it to length) to have an indicator about the maximum length. Then we add a sample counter. Or we can use `write_index` for it?

Sounds good, or even max_samples or sample_buffer_size so it's clear it's a different length

app/src/history/zsw_history.c
Kampi: Okay so Ieave it as it is.

jakkra: > Okay so Ieave it as it is.

No remove

app/src/history/zsw_history.c
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    return 0;
+}
+
+int zsw_history_samples(zsw_history_t *p_history)
+{
+    __ASSERT(p_history != NULL, "Invalid parameters for zsw_history_samples");
+
+    return p_history->write_index;
```
jakkra, line 206: Will not work, when max samples are used, the write_index will wrap around to 0 again.
Need to add another variable into the setting struct. Adding that.

app/src/history/zsw_history.c
```diff
@@ -0,0 +1,207 @@
...
+    LOG_DBG("Load data with key %s", key_data);
+    if (error) {
+        LOG_ERR("Error during data loading! Error: %i", error);
+        return -EFAULT;
+    }
+
+    return 0;
+}
+
+int zsw_history_save(zsw_history_t *p_history, const void *p_sample)
```
jakkra, line 175: p_sample unused, causing battery_app not to add any samples. I'm fixing this by adding the sample if p_sample != NULL

jakkra (changes requested):

app/src/history/zsw_history.c
Kampi: No - `zsw_history_save` should only save the history to NVS without adding a sample. I just forgot to remove this parameter yesterday.

Kampi: The workflow should be to use `zsw_history_add` to add the samples to the history and then call `zsw_history_save` when you are done and when you want to save it.

jakkra:
Ok, it's my voltage compressions int mV => byte which is bad, nothing wrong with history code I think.

jakkra (approved):
Done and I think it works.

PR #268, opened 2024-05-08
Add missing offsets for month and year

Kampi:


app/src/ui/watchfaces/107_2_dial/zsw_watchface_107_2_dial_ui.c
```diff
@@ -190,22 +190,22 @@ static void watchface_107_2_dial_set_datetime(int day_of_week, int date, int day
 
     if (getPlaceValue(last_month, 1) != getPlaceValue(month, 1)) {
```
jakkra, line 191: Need to change all I think, otherwise the last value check will not work. So `getPlaceValue` should also match day here and month on the other

app/src/zsw_clock.c
```diff
@@ -45,6 +45,13 @@ void zsw_clock_get_time(zsw_timeval_t *ztm)
     gettimeofday(&tv, NULL);
     tm = localtime(&tv.tv_sec);
     memcpy(ztm, tm, sizeof(struct tm));
+
+    // Add one to the month because we want to count from December instead of January
+    ztm->tm.tm_mon += 1;
```
jakkra, line 50: This will break as tm_mon and tm_year are already increased at other places in the code. Also ble_cts.c adjusts the time to match current implementation. I suggest change it in the watchface_app -> setDateTime instead.

jakkra (changes requested):

app/src/zsw_clock.c
Kampi: mmh or shall we remove it there and do it in one place (zsw_clock)? I don´t like the idea that the user has to take care about it and must add this every time.

Kampi: Ok so I check the code for other places and change them too.

jakkra (approved):

PR #297, opened 2024-06-19
Update checkout CI version

ldab:
v3 is deprecated


Kampi:
`actions/upload-artifact@v3` is depreached too.

PR #305, opened 2024-07-12
rev 5: Final Flash config.

jakkra:
Mactronix 64MB flash working.

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -27,8 +27,16 @@ if("nrf5340dk_nrf5340_cpuapp" STREQUAL "${BOARD}")
 endif()
 
 set(BOARD_ROOT ${CMAKE_CURRENT_SOURCE_DIR})
+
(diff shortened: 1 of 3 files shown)


app/CMakeLists.txt
```diff
@@ -27,9 +27,17 @@ if("nrf5340dk_nrf5340_cpuapp" STREQUAL "${BOARD}")
...
 
 set(BOARD_ROOT ${CMAKE_CURRENT_SOURCE_DIR})
+
 find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
 project(ZSWatchFW)
 
+# These are XIP related anomalies and aren't applicable for us and cause
+# throughput issues and QSPI issues on NCS 2.6.0
+zephyr_compile_definitions(
+  -DNRF53_ERRATA_43_ENABLE_WORKAROUND=0
```
ldab, line 37: 🫨

jakkra: Not needed in NCS 2.7, but we will se about this change then, keep or remove.

PR #315, opened 2024-07-17
Print INFO log on green color on RTT

ldab:
Yay? I like how it looks, but it comes with a cost


jakkra:
Like it! Just maybe add it to debug.conf instead?

Kampi:
> Yay? I like how it looks, but it comes with a cost

Shall we add red for errors too?

ldab:
> > Yay? I like how it looks, but it comes with a cost
> 
> Shall we add red for errors too?

That is already there, warning are yellow

PR #322, opened 2024-07-20
spi_nor: do not DPD in idle on HW3

ldab:
As CI will show, by changing this the overall performance increases significantly


Kampi:
Is there a way to manually set the flash in DPD? Otherwise we will have an increased power consumption

jakkra:
> Is there a way to manually set the flash in DPD? Otherwise we will have an increased power consumption

I'm not sure exactly what this config does, but if it means never dpd, the cut is from 2uA to 10uA which is not that bad overall. This also only affects v3 pcb.

Manual enter and exit of dpd I'll investigate on v5, as doing it while we are active will just slow down.

Kampi:
mmh let´s test it :)
I´m not sure about the current consumption yet, because the current consumption isn´t good, so I would avoid adding more.

jakkra:
> mmh let´s test it :) I´m not sure about the current consumption yet, because the current consumption isn´t good (with my rev4), so I would avoid adding more.

Does does not affect v4, it's v3 only because this is for spi driver and not the QSPI driver

Kampi:
> > mmh let´s test it :) I´m not sure about the current consumption yet, because the current consumption isn´t good (with my rev4), so I would avoid adding more.
> 
> Does does not affect v4, it's v3 only because this is for spi driver and not the QSPI driver

Ah okay. Guess I understand it wrong :)

jakkra (approved):

PR #327, opened 2024-07-24
HW test optimizations

Kampi:
- Pass J-Link serial and hardware revision as arguments to pytest
- Add a setup for different hardware revisions and J-Links to CI/CD
- Add support for ARM 64
- Add support for PPK2
- Skip the test when the target J-Link isn´t connected
- Device recovering
- Test report

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,12 +2,16 @@ on: [workflow_call]
 
 jobs:
   test:
-    runs-on: self-hosted
+    runs-on: [self-hosted, Linux, ZSWatch]
     container: 
       image: ghcr.io/zephyrproject-rtos/zephyr-build:v0.26.2
       volumes:
         - /dev/bus/usb/:/dev/bus/usb
       options: --privileged
+
+    strategy:
+      matrix:
+         devices: [{hardware_revision: 5, jlink_device: 760208506}, {hardware_revision: 3, jlink_device: 760208490}, {hardware_revision: 3, jlink_device: 760208484}]
 
     steps:
       - uses: actions/checkout@v4
@@ -20,32 +24,52 @@ jobs:
       - name: Install dependencies
         continue-on-error: true # JLink install will fail
         run: |
-          if [ ! -f "`which JLinkExe`" ]; then
-            echo "Installing JLink tools..."
-            wget -q --post-data 'accept_license_agreement=accepted&non_emb_ctr=confirmed&submit=Download+software' https://www.segger.com/downloads/jlink/JLink_Linux_V796t_x86_64.deb
-            sudo apt -qq update
-            sudo apt -y install ./JLink_Linux_V796t_x86_64.deb &>/dev/null
+          arch=$(arch)
+          echo "Detected arch: ${arch}"
+          if [ "${arch}" = "x86_64" ]; then
+              echo "x64 Architecture"
+              JLINK_NAME=JLink_Linux_V796t_x86_64
+              NRF_NAME=nrf-command-line-tools_10.24.2_amd64
+          elif [ "${arch}" = "aarch64" ]; then
+              echo "ARM 64 Architecture"
+              JLINK_NAME=JLink_Linux_V796t_arm
+              NRF_NAME=nrf-command-line-tools_10.24.2_arm64
+              sudo dpkg --add-architecture armhf
           fi
 
-          if [ ! -f "`which nrfjprog`" ]; then
-            echo "Installing Nordic command line tools..."
-            wget -q https://nsscprodmedia.blob.core.windows.net/prod/software-and-other-downloads/desktop-software/nrf-command-line-tools/sw/versions-10-x-x/10-24-2/nrf-command-line-tools_10.24.2_amd64.deb
-            sudo apt -y install ./nrf-command-line-tools_10.24.2_amd64.deb &>/dev/null
-            export PATH=$PATH:/opt
-          fi
+          sudo apt-get -qq update
+
+          echo "Installing JLink tools..."
+          wget -q --post-data 'accept_license_agreement=accepted&non_emb_ctr=confirmed&submit=Download+software' https://www.segger.com/downloads/jlink/${JLINK_NAME}.deb
+          sudo dpkg --unpack ./${JLINK_NAME}.deb
+          sudo rm /var/lib/dpkg/info/jlink.postinst -f
+          sudo dpkg --configure jlink
+          sudo apt install -yf 
+
+          echo "Installing Nordic command line tools..."
+          wget -q https://nsscprodmedia.blob.core.windows.net/prod/software-and-other-downloads/desktop-software/nrf-command-line-tools/sw/versions-10-x-x/10-24-2/${NRF_NAME}.deb
+          sudo apt-get -y install ./${NRF_NAME}.deb
 
           python -m pip install --upgrade pip
           pip install -r app/pytest/requirements.txt
 
-      - name: 'Download image'
+          nrfjprog -i
+
+      - name: Download debug image
         uses: actions/download-artifact@v4
         with:
-          name: zswatch_nrf5340_cpuapp@3_debug
-
-      - name: Display structure of downloaded files
-        run: ls
+          name: zswatch_nrf5340_cpuapp@${{ matrix.devices.hardware_revision }}_debug
 
       - name: Test with pytest
+        continue-on-error: true # Some tests can fail
         run: |
           pip install pytest
-          pytest app/pytest/
+          pytest --html=report/report.html --jlink=${{ matrix.devices.jlink_device }} --hw=${{ matrix.devices.hardware_revision }} app/pytest/
+
+      - name : Publish results
+        uses: actions/upload-artifact@v4.3.3
+        with:
+          name: ${{ matrix.devices.jlink_device }}_${{ matrix.devices.hardware_revision }}
+          path: |
+            report
+          if-no-files-found: ignore
--- a/app/pytest/utils.py
+++ b/app/pytest/utils.py
@@ -3,17 +3,15 @@ import logging
 import time
 import pylink
 
-SERIAL_NUMBER = "760208490"
+from ppk2_api.ppk2_api import PPK2_API
 
 log = logging.getLogger()
-
 
 def current_milli_time():
     """Get times in milliseconds"""
     return round(time.time() * 1000)
 
-
-def reset():
+def reset(serial):
     """Reset DUT"""
     log.info("Reset DUT")
     subprocess.run(
@@ -22,7 +20,7 @@ def reset():
             "--family",
             "nrf53",
             "--snr",
-            SERIAL_NUMBER,
+            serial,
             "--reset",
         ],
         shell=False,
@@ -31,8 +29,7 @@ def reset():
         check=True,
     )
 
-
-def flash():
+def flash(serial, hw):
     """Flash firmware"""
     log.info("Flashing CP_APPLICATION.")
     subprocess.run(
@@ -41,9 +38,9 @@ def flash():
             "--family",
             "nrf53",
             "--snr",
-            SERIAL_NUMBER,
+            serial,
             "--program",
-            "./zswatch_nrf5340_cpuapp@3_debug.hex",
+            "./zswatch_nrf5340_cpuapp@{}_debug.hex".format(hw),
             "--chiperase",
             "--verify",
         ],
@@ -60,7 +57,7 @@ def flash():
             "--family",
             "nrf53",
             "--snr",
-            SERIAL_NUMBER,
+            serial,
             "--program",
             "./zswatch_nrf5340_CPUNET.hex",
             "--coprocessor",
@@ -75,13 +72,12 @@ def flash():
         check=True,
     )
 
-
-def read_rtt(target_device="nRF5340_XXAA", timeout_ms=10000):
+def read_rtt(serial, target_device="nRF5340_XXAA", timeout_ms=10000):
     """Read Segger RTT output"""
     jlink = pylink.JLink()
     logging.getLogger("pylink.jlink").setLevel(logging.WARNING)
     log.info("Connecting to JLink...")
-    jlink.open(serial_no=SERIAL_NUMBER)
+    jlink.open(serial_no=serial)
     log.info("Connecting to %s..." % target_device)
     jlink.set_tif(pylink.enums.JLinkInterfaces.SWD)
     jlink.connect(target_device)
@@ -103,3 +99,41 @@ def read_rtt(target_device="nRF5340_XXAA", timeout_ms=10000):
     log.debug(read_data)
 
     return read_data
+
+def ppk_connected():
+    return PPK2_API.list_devices()
+
+def is_locked(serial, target_device="nRF5340_XXAA"):
+    '''Check if the target is locked'''
+    
+    locked = False
+    
+    jlink = pylink.JLink()
+    log.info("Connecting to JLink...")
+    jlink.open(serial_no=serial)
+    log.info("Connecting to %s..." % target_device)
+    jlink.set_tif(pylink.enums.JLinkInterfaces.SWD)
+    try:
+        jlink.connect(target_device)
+    except pylink.errors.JLinkException as e:
+        locked = True
+    finally:
+        jlink.close()
+
+    return locked
+
+def unlock():
+    unlocked = False
+
+    subprocess.run(
+        [
+            "nrfjprog",
+            "--recover"
+        ],
+        shell=False,
+        stderr=subprocess.STDOUT,
+        text=True,
+        check=True,
+    )
+
+    return unlocked
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,5 +7,5 @@ jobs:
     uses: ./.github/workflows/build.yml
 
   test:
-    needs: [build]
+    #needs: [build]
     uses: ./.github/workflows/test.yml
--- a/app/pytest/assets/style.css
+++ b/app/pytest/assets/style.css
@@ -1,0 +1,319 @@
+body {
+  font-family: Helvetica, Arial, sans-serif;
+  font-size: 12px;
+  /* do not increase min-width as some may use split screens */
+  min-width: 800px;
+  color: #999;
+}
+
+h1 {
+  font-size: 24px;
+  color: black;
+}
+
+h2 {
+  font-size: 16px;
+  color: black;
+}
+
+p {
+  color: black;
+}
+
+a {
+  color: #999;
+}
+
+table {
+  border-collapse: collapse;
+}
+
+/******************************
+ * SUMMARY INFORMATION
+ ******************************/
+#environment td {
+  padding: 5px;
+  border: 1px solid #e6e6e6;
+  vertical-align: top;
+}
+#environment tr:nth-child(odd) {
+  background-color: #f6f6f6;
+}
+#environment ul {
+  margin: 0;
+  padding: 0 20px;
+}
+
+/******************************
+ * TEST RESULT COLORS
+ ******************************/
+span.passed,
+.passed .col-result {
+  color: green;
+}
+
+span.skipped,
+span.xfailed,
+span.rerun,
+.skipped .col-result,
+.xfailed .col-result,
+.rerun .col-result {
+  color: orange;
+}
+
+span.error,
+span.failed,
+span.xpassed,
+.error .col-result,
+.failed .col-result,
+.xpassed .col-result {
+  color: red;
+}
+
+.col-links__extra {
+  margin-right: 3px;
+}
+
+/******************************
+ * RESULTS TABLE
+ *
+ * 1. Table Layout
+ * 2. Extra
+ * 3. Sorting items
+ *
+ ******************************/
+/*------------------
+ * 1. Table Layout
+ *------------------*/
+#results-table {
+  border: 1px solid #e6e6e6;
+  color: #999;
+  font-size: 12px;
+  width: 100%;
+}
+#results-table th,
+#results-table td {
+  padding: 5px;
+  border: 1px solid #e6e6e6;
+  text-align: left;
+}
+#results-table th {
+  font-weight: bold;
+}
+
+/*------------------
+ * 2. Extra
+ *------------------*/
+.logwrapper {
+  max-height: 230px;
+  overflow-y: scroll;
+  background-color: #e6e6e6;
+}
+.logwrapper.expanded {
+  max-height: none;
+}
+.logwrapper.expanded .logexpander:after {
+  content: "collapse [-]";
+}
+.logwrapper .logexpander {
+  z-index: 1;
+  position: sticky;
+  top: 10px;
+  width: max-content;
+  border: 1px solid;
+  border-radius: 3px;
+  padding: 5px 7px;
+  margin: 10px 0 10px calc(100% - 80px);
+  cursor: pointer;
+  background-color: #e6e6e6;
+}
+.logwrapper .logexpander:after {
+  content: "expand [+]";
+}
+.logwrapper .logexpander:hover {
+  color: #000;
+  border-color: #000;
+}
+.logwrapper .log {
+  min-height: 40px;
+  position: relative;
+  top: -50px;
+  height: calc(100% + 50px);
+  border: 1px solid #e6e6e6;
+  color: black;
+  display: block;
+  font-family: "Courier New", Courier, monospace;
+  padding: 5px;
+  padding-right: 80px;
+  white-space: pre-wrap;
+}
+
+div.media {
+  border: 1px solid #e6e6e6;
+  float: right;
+  height: 240px;
+  margin: 0 5px;
+  overflow: hidden;
+  width: 320px;
+}
+
+.media-container {
+  display: grid;
+  grid-template-columns: 25px auto 25px;
+  align-items: center;
+  flex: 1 1;
+  overflow: hidden;
+  height: 200px;
+}
+
+.media-container--fullscreen {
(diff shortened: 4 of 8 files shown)


app/pytest/utils.py
```diff
@@ -33,41 +32,20 @@ def reset():
...
+    log.info("Flashing application.")
     subprocess.run(
         [
-            "nrfjprog",
-            "--family",
-            "nrf53",
+            "west",
+            "flash",
+            "-d",
+            "/home/daniel/zephyr_projects/ZSWatch/app/" + os.environ["ZSWATCH_BUILD_DIR"],
```
ldab, line 44: The path seems wrong

app/pytest/pytest.ini
```diff
@@ -1,3 +1,6 @@
 [pytest]
 log_cli = True
 log_cli_level = INFO
+env =
```
ldab, line 4: I think the idea is to have the serial number to be passed as an argument of the CI, like a matrix sort of thing, that way multiple can be used

app/pytest/pytest.ini
```diff
@@ -1,3 +1,6 @@
 [pytest]
 log_cli = True
 log_cli_level = INFO
+env = 
+    JLINK_SERIAL=760208506
+    ZSWATCH_BUILD_DIR=debug_rev5
```
ldab, line 6: The west flash will not work because this job runs on a different computer than the one that builds the fw

app/pytest/utils.py
Kampi: Hahaha yes. Copy + Paste error :P

app/pytest/pytest.ini
Kampi: You mean like this? https://stackoverflow.com/a/37396406

It depends on the use case. Both makes sense

app/pytest/pytest.ini
Kampi: Can you explain it? So you want to split the setup into two PCs. One for developing firmware without HW access and one for the testing with HW access?

jakkra: It's already like that. As it's super slow to run all build locally. The HW test job will download the built FW from github and flash it then run test.

jakkra: See here https://github.com/jakkra/ZSWatch/blob/d540a61c8f03c414f9232618d87866d10c5d7987/.github/workflows/test.yml#L40

Kampi: Okay guess I understand the idea behind it wrong. But then we need a download based on the HW revision or we need to pass the HW revision into the test so the correct image is downloaded.

jakkra: Correct this is needed. So we need jlink ID and it's version.

app/pytest/pytest.ini
ldab: More like this ‘cmdopts’ https://docs.pytest.org/en/latest/example/simple.html

so the yml job file call pytest with the argument, ther argument is HW version and serial number

app/pytest/pytest.ini
ldab: Build job https://github.com/jakkra/ZSWatch/blob/d540a61c8f03c414f9232618d87866d10c5d7987/.github/workflows/build.yml#L5

test job https://github.com/jakkra/ZSWatch/blob/d540a61c8f03c414f9232618d87866d10c5d7987/.github/workflows/test.yml#L5

app/pytest/pytest.ini
```diff
@@ -1,3 +1,6 @@
 [pytest]
 log_cli = True
 log_cli_level = INFO
+env = 
+    JLINK_SERIAL=760208506
+    ZSWATCH_REVISION=5
```
ldab, line 6: Btw, this serial number uses hw3

app/pytest/pytest.ini
Kampi: So we can use `JLINK_SERIAL=760208506 ZSWATCH_REVISION=5 pytest -c pytest/pytest.ini` to pass arguments like the serial number and the hardware revision into the test (i. e. from CI/CD). All we need now is a way to pass these parameter into the CI/CD pipeline

ldab:
So the way I imagined it is like

1. yml file has a matrix with Self-runner name + HW version + Serial
2. Pytest …. $matrix

The self hosted runner name is important on a matrix because I have one now, Jakob may set up another one and so on…. And maybe we need another argument with features if you set up one with a PowerProfiler 🤯 so that not all tests would apply anymore

app/pytest/pytest.ini
Kampi: Yeah it´s for me for testing. We can change it back later.

app/pytest/pytest.ini
Kampi: > Build job
> 
> https://github.com/jakkra/ZSWatch/blob/d540a61c8f03c414f9232618d87866d10c5d7987/.github/workflows/build.yml#L5
> 
> test job
> 
> https://github.com/jakkra/ZSWatch/blob/d540a61c8f03c414f9232618d87866d10c5d7987/.github/workflows/test.yml#L5

I´m not an expert with CI/CD. Can I run this workflow locally? And how can I do it? Do you have a good manual for it?

app/pytest/pytest.ini
ldab: And need to link with https://github.com/jakkra/ZSWatch/blob/d540a61c8f03c414f9232618d87866d10c5d7987/.github/workflows/test.yml#L43

app/pytest/pytest.ini
ldab: It is a pain in the ass to run the whole CI locally, i personally never managed and aways gave up when i tried hahahaha

ldab: > So we can use `JLINK_SERIAL=760208506 ZSWATCH_REVISION=5 pytest -c pytest/pytest.ini` to pass arguments like the serial number and the hardware revision into the test (i. e. from CI/CD). All we need now is a way to pass these parameter into the CI/CD pipeline

Kind of… the ‘.ini’ file is automatically selected so don’t need to pass as argument, so here

https://github.com/jakkra/ZSWatch/blob/d540a61c8f03c414f9232618d87866d10c5d7987/.github/workflows/test.yml#L49

I’d change to ‘pytest app/pytest/ SERIAL VERSION’

Kampi:
> So the way I imagined it is like
> 
> 1. yml file has a matrix with Self-runner name + HW version + Serial
> 2. Pytest …. $matrix
> 
> The self hosted runner name is important on a matrix because I have one now, Jakob may set up another one and so on…. And maybe we need another argument with features if you set up one with a PowerProfiler 🤯 so that not all tests would apply anymore



> So the way I imagined it is like
> 
> 1. yml file has a matrix with Self-runner name + HW version + Serial
> 2. Pytest …. $matrix
> 
> The self hosted runner name is important on a matrix because I have one now, Jakob may set up another one and so on…. And maybe we need another argument with features if you set up one with a PowerProfiler 🤯 so that not all tests would apply anymore

I added two matrices for the J-Links and the HW revisions so we can start in a simple way with it. Ofc we need a way to pass options because maybe you also want to compare `debug` and `release` build too.

ldab:
> > So the way I imagined it is like
> > 
> > 1. yml file has a matrix with Self-runner name + HW version + Serial
> > 2. Pytest …. $matrix
> > 
> > The self hosted runner name is important on a matrix because I have one now, Jakob may set up another one and so on…. And maybe we need another argument with features if you set up one with a PowerProfiler 🤯 so that not all tests would apply anymore
> 
> > So the way I imagined it is like
> > 
> > 1. yml file has a matrix with Self-runner name + HW version + Serial
> > 2. Pytest …. $matrix
> > 
> > The self hosted runner name is important on a matrix because I have one now, Jakob may set up another one and so on…. And maybe we need another argument with features if you set up one with a PowerProfiler 🤯 so that not all tests would apply anymore
> 
> I added two matrices for the J-Links and the HW revisions so we can start in a simple way with it. Ofc we need a way to pass options because maybe you also want to compare `debug` and `release` build too.

Seem right, @jakkra I thought computer name can be passed as argument but apparently only labels https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/using-self-hosted-runners-in-a-workflow#about-self-hosted-runner-labels can you please check what labels are available on this runner?

Kampi:
> So the way I imagined it is like
> 
> 1. yml file has a matrix with Self-runner name + HW version + Serial
> 2. Pytest …. $matrix
> 
> The self hosted runner name is important on a matrix because I have one now, Jakob may set up another one and so on…. And maybe we need another argument with features if you set up one with a PowerProfiler 🤯 so that not all tests would apply anymore

The command line option feels like the better solution. But then we need a way to pass the variables from `conftest` to `test_basic` because `test_basic` is also using the `utils`

ldab:
> > So the way I imagined it is like
> > 
> > 1. yml file has a matrix with Self-runner name + HW version + Serial
> > 2. Pytest …. $matrix
> > 
> > The self hosted runner name is important on a matrix because I have one now, Jakob may set up another one and so on…. And maybe we need another argument with features if you set up one with a PowerProfiler 🤯 so that not all tests would apply anymore
> 
> The command line option feels like the better solution. But then we need a way to pass the variables from `conftest` to `test_basic` because `test_basic` is also using the `utils`

Need a fixture, I believe the second example here should do it https://docs.pytest.org/en/latest/example/simple.html

.github/workflows/test.yml
```diff
@@ -9,6 +9,12 @@ jobs:
         - /dev/bus/usb/:/dev/bus/usb
       options: --privileged
 
+    strategy:
+      matrix:
+        hardware_revisions: [3, 4]
+        jlink_devices: [760208506]
+        runner: []
```
ldab, line 16: This can’t be empty, for now you can use “ self-hosted”

and change to ‘runs-on: ${{ matrix.runner }}’

Kampi:
> > > So the way I imagined it is like
> > > 
> > > 1. yml file has a matrix with Self-runner name + HW version + Serial
> > > 2. Pytest …. $matrix
> > > 
> > > The self hosted runner name is important on a matrix because I have one now, Jakob may set up another one and so on…. And maybe we need another argument with features if you set up one with a PowerProfiler 🤯 so that not all tests would apply anymore
> > 
> > 
> > The command line option feels like the better solution. But then we need a way to pass the variables from `conftest` to `test_basic` because `test_basic` is also using the `utils`
> 
> Need a fixture, I believe the second example here should do it https://docs.pytest.org/en/latest/example/simple.html

Added both parameters as command line arguments 👍

Kampi:
@ldab how can I add my J-Link to the runner?

image

It´s connected to the system but the runner can´t find it

image

ldab:
> @ldab how can I add my J-Link to the runner?
> 
> image
> 
> It´s connected to the system but the runner can´t find it
> 
> image

I don’t know if that is possible. My jlink is connect to my computer (aka runner) so it’s actually real HW connected physically together

Kampi:
> > > So the way I imagined it is like
> > > 
> > > 1. yml file has a matrix with Self-runner name + HW version + Serial
> > > 2. Pytest …. $matrix
> > > 
> > > The self hosted runner name is important on a matrix because I have one now, Jakob may set up another one and so on…. And maybe we need another argument with features if you set up one with a PowerProfiler 🤯 so that not all tests would apply anymore
> > 
> > 
> > > So the way I imagined it is like
> > > 
> > > 1. yml file has a matrix with Self-runner name + HW version + Serial
> > > 2. Pytest …. $matrix
> > > 
> > > The self hosted runner name is important on a matrix because I have one now, Jakob may set up another one and so on…. And maybe we need another argument with features if you set up one with a PowerProfiler 🤯 so that not all tests would apply anymore
> > 
> > 
> > I added two matrices for the J-Links and the HW revisions so we can start in a simple way with it. Ofc we need a way to pass options because maybe you also want to compare `debug` and `release` build too.
> 
> Seem right, @jakkra I thought computer name can be passed as argument but apparently only labels https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/using-self-hosted-runners-in-a-workflow#about-self-hosted-runner-labels can you please check what labels are available on this runner?

I set up a runner for me with these labels

image


> > @ldab how can I add my J-Link to the runner?
> > image
> > It´s connected to the system but the runner can´t find it
> > image
> 
> I don’t know if that is possible. My jlink is connect to my computer (aka runner) so it’s actually real HW connected physically together

My runner runs inside a Ubuntu VM and my J-Link is connected to this VM.

ldab:
> > > @ldab how can I add my J-Link to the runner?
> > > image
> > > It´s connected to the system but the runner can´t find it
> > > image
> > 
> > 
> > I don’t know if that is possible. My jlink is connect to my computer (aka runner) so it’s actually real HW connected physically together
> 
> My runner runs inside a Ubuntu VM and my J-Link is connected to this VM.

I’m not sure if that is the case but when starting the docker you need “privileged” and https://github.com/jakkra/ZSWatch/blob/d540a61c8f03c414f9232618d87866d10c5d7987/.github/workflows/test.yml#L9

Kampi:
> I’m not sure if that is the case but when starting the docker you need “privileged” and
> 
> https://github.com/jakkra/ZSWatch/blob/d540a61c8f03c414f9232618d87866d10c5d7987/.github/workflows/test.yml#L9

Ah okay you run it via Docker? I use the `run.sh` script from the GitHub instructions. Let me check how to change it.

ldab:
> > I’m not sure if that is the case but when starting the docker you need “privileged” and
> > https://github.com/jakkra/ZSWatch/blob/d540a61c8f03c414f9232618d87866d10c5d7987/.github/workflows/test.yml#L9
> 
> Ah okay you run it via Docker? I use the `run.sh` script from the GitHub instructions. Let me check how to change it.

I run mine as a service… but I thought you were running a docker locally.

if running the CI yml those arguments I mentioned are already there

ldab:
> @ldab how can I add my J-Link to the runner?
> 
> image
> 
> It´s connected to the system but the runner can´t find it
> 
> image

Bear in mind that this screenshot is probably from Jakkra runner.

And the job only runs on PR or manually..

so you need to go to your repo and trigger the job from your CI.yml new workflow or something like that, otherwise jakkra repo does not have access to your runner (you need to ask him for a token if you want that)

Kampi:
> > @ldab how can I add my J-Link to the runner?
> > image
> > It´s connected to the system but the runner can´t find it
> > image
> 
> Bear in mind that this screenshot is probably from Jakkra runner.
> 
> And the job only runs on PR or manually..
> 
> so you need to go to your repo and trigger the job from your CI.yml new workflow or something like that, otherwise jakkra repo does not have access to your runner (you need to ask him for a token if you want that)

Ah! Good catch. That explains why my labels don´t work.

Kampi:
Got this when running it in my repo

image

Do you have an idea?

Kampi:
@ldab it´s running now. You have to run

```
sudo apt install docker.io
sudo chmod 666 /var/run/docker.sock
```

in your machine to make it runable.

ldab:
> @ldab it´s running now. You have to run
> 
> ```
> sudo apt install docker.io
> sudo chmod 666 /var/run/docker.sock
> ```
> 
> in your machine to make it runable.

I probably did that at some stage on mine also 😹

Kampi:
> > in your machine to make it runable.
> 
> I probably did that at some stage on mine also 😹

Maybe I will work at some time for me. Now I´m running out of space in my VM...

ldab:
> > > in your machine to make it runable.
> > 
> > 
> > I probably did that at some stage on mine also 😹
> 
> Maybe I will work at some time for me. Now I´m running out of space in my VM...

Time to go shopping then

ldab:
@Kampi the job seems stuck, change the runs on to “self hosted” only as it may be ANDing the labels that may not be there

Kampi:
> @Kampi the job seems stuck, change the runs on to “self hosted” only as it may be ANDing the labels that may not be there

Yeah I played around with it. I will change it back later.

Kampi:
Yeah it´s working 💯 (except handling the device protection)

https://github.com/Kampi/ZSWatch/actions/runs/10074225612/job/27850275408

ldab:
> Yeah it´s working 💯 (except handling the device protection)
> 
> https://github.com/Kampi/ZSWatch/actions/runs/10074225612/job/27850275408

One of those one liners PR that grows right? Hahaha

the link you sent is not valid.

I was thinking, why don’t we enquire the PC for what Jlink it has and then check against a lookup array?

Kampi:
> One of those one liners PR that grows right? Hahaha
> 

True. It escalated quickly :D

> the link you sent is not valid.

Yeah because I made some changes to fix some typos in the Python script. Looks better now :)

> I was thinking, why don’t we enquire the PC for what Jlink it has and then check against a lookup array?

mmh I think it´s not a good way, because someone (for example me) can use different HW platforms and only one J-Link. I would prefere a solution like writing the HW revision into a fixed place in memory, read this place with the J-Link and then use this version for the test.

ldab:
> I would prefere a solution like writing the HW revision into a fixed place in memory,

That is already there no ? There is some defines, not sure where those are stored

otherwise my experience points to some linker file voodoo

Kampi:
> > I would prefere a solution like writing the HW revision into a fixed place in memory,
> 
> That is already there no ? There is some defines, not sure where those are stored

The defines (i. e. `CONFIG_ZSWATCH_PCB_REV`) are included somewhere in the code. Not sure if the place of the value in the compiled binary is always the same, but I think it can differ between different versions (because the code is growing).

> otherwise my experience points to some linker file voodoo

You can reduce the maximum binary size by one word and fill this word with the HW revision. I did it ~10 years ago for an ATTiny to store a serial number in the binary for the flash memory.

Kampi:
@ldab the idea is stupid. We need the hardware version to download the right firmware so we know it already. If we don´t want to pass the version we need something to identify the hardware without a firmware.

ldab:
> @ldab the idea is stupid :D. We need the hardware version to download the right firmware so we know it already. If we don´t want to pass the version we need something to identify the hardware without a firmware.

Not really, you can create a step before to download to “identify hw”

let me look at the linker file, i did at work recently, it’s fresh-ish but let’s leave for another pr

Kampi:
> > @ldab how can I add my J-Link to the runner?
> > image
> > It´s connected to the system but the runner can´t find it
> > image
> 
> Bear in mind that this screenshot is probably from Jakkra runner.
> 
> And the job only runs on PR or manually..
> 
> so you need to go to your repo and trigger the job from your CI.yml new workflow or something like that, otherwise jakkra repo does not have access to your runner (you need to ask him for a token if you want that)



> > @ldab the idea is stupid :D. We need the hardware version to download the right firmware so we know it already. If we don´t want to pass the version we need something to identify the hardware without a firmware.
> 
> Not really, you can create a step before to download to “identify hw”
> 
> let me look at the linker file, i did at work recently, it’s fresh-ish but let’s leave for another pr

Yeah but how do you want to identify the hw when no firmware is flashed? Or you have to use a generic firmware for all variants which is flashed at first. This firmware identifies the hardware version, report it back to the test and then the test is flashes the right firmware.

ldab:
> Yeah but how do you want to identify the hw when no firmware is flashed?

Good point, I guess one could assume that some firmware would be flashed... but I guess it can be a topic for another day, hardcodding now seems fine

jakkra:
@
Screenshot_2024-07-24-16-27-22-35_40deb401b9ffe8e1df2f1cc5ba480b12
Can add any label wanted.

jakkra:
Nice progress by the way, very cool 😀

> > Not really, you can create a step before to download to “identify hw”
> > let me look at the linker file, i did at work recently, it’s fresh-ish but let’s leave for another pr
> 
> Yeah but how do you want to identify the hw when no firmware is flashed? Or you have to use a generic firmware for all variants which is flashed at first. This firmware identifies the hardware version, report it back to the test and then the test is flashes the right firmware.

Don't think this need to be generic, each jlink ID will be connected to a specific hw version, so it's in the matrix/config file. If this change, then the config/matrix is changed.

jakkra:
> > I would prefere a solution like writing the HW revision into a fixed place in memory,
> 
> That is already there no ? There is some defines, not sure where those are stored
> 
> otherwise my experience points to some linker file voodoo

We can use UICR to write the FW version once, but this will be removed if a - - recover or ereseall command is used when flashing. So need to make sure not to do that then.

ldab:
PS. My runner is down, grub problems

.github/workflows/test.yml
```diff
@@ -2,13 +2,17 @@ on: [workflow_call]
...
     container: 
       image: ghcr.io/zephyrproject-rtos/zephyr-build:v0.26.2
       volumes:
         - /dev/bus/usb/:/dev/bus/usb
       options: --privileged
 
+    strategy:
+      matrix:
+         devices: [{hardware_revision: 5, jlink_device: 760208506}, {hardware_revision: 3, jlink_device: 760208507},]
+
```
ldab, line 15: My serial is 760208490

Kampi:
@jakkra @ldab we need a way to handle something like a timeout when the job is waiting for a runner, otherwise all jobs will wait until a runner becomes active.

jakkra:
> @jakkra @ldab we need a way to handle something like a timeout when the job is waiting for a runner, otherwise all jobs will wait until a runner becomes active.

Why is it a problem? You want to silently fail a pipeline if other pipelines are running and queue is long?

ldab:
> @jakkra @ldab we need a way to handle something like a timeout when the job is waiting for a runner, otherwise all jobs will wait until a runner becomes active.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

.github/workflows/test.yml
```diff
@@ -2,13 +2,17 @@ on: [workflow_call]
...
+    runs-on: [self-hosted]
     container: 
       image: ghcr.io/zephyrproject-rtos/zephyr-build:v0.26.2
       volumes:
         - /dev/bus/usb/:/dev/bus/usb
       options: --privileged
 
+    strategy:
+      matrix:
+         devices: [{hardware_revision: 5, jlink_device: 760208506}, {hardware_revision: 3, jlink_device: 760208490},]
```
ldab, line 14: @jakkra please add some meaningful tag to both runners, something like 760208490_3
than we can use the tag on “runs on”, what do you think @Kampi ?

Kampi: Sounds good.

Kampi:
Added support for ARM 64 bit so you can use a Raspberry Pi 4 as runner

ldab:
> Added support for ARM 64-Bit so you can use a Raspberry Pi 4 as a runner

Cool, I’m afraid something is not right tho https://github.com/jakkra/ZSWatch/actions/runs/10088096496/job/27893855338?pr=327#step:5:42

Kampi:
@ldab had the issue that these packages couldn´t get installed from the runner.

image

Installing JLink directly on the Raspberry Pi was working.

Kampi:
> > Added support for ARM 64-Bit so you can use a Raspberry Pi 4 as a runner
> 
> Cool, I’m afraid something is not right tho https://github.com/jakkra/ZSWatch/actions/runs/10088096496/job/27893855338?pr=327#step:5:42

It´s a typo in `test.yml`

```
NRF_Nam${arch}f-command-line-tools_10.24.2_amd64
```

should be 

```
NRF_Name=${arch}f-command-line-tools_10.24.2_amd64
```

.github/workflows/test.yml
```diff
@@ -20,32 +24,42 @@ jobs:
       - name: Install dependencies
         continue-on-error: true # JLink install will fail
         run: |
+          arch=$(arch)
+          echo "Detected arch: ${arch}"
+          if [ "${arch}" = "x86_64" ]; then
+              echo "x64 Architecture"
+              JLINK_Name=JLink_Linux_V796t_x86_64
+              NRF_Nam${arch}f-command-line-tools_10.24.2_amd64
```
ldab, line 32:
```suggestion
              NRF_Name=nrf-command-line-tools_10.24.2_amd64
```

Kampi: Too slow :P

Kampi:
I have some DLL errors with JLink on my Raspberry Pi. Need to check it later. Maybe the installation wasn´t correct or I missed something. A PPK2 is also connected to the Raspberry Pi (but not to the watch yet) so we can try to play around with it.

ldab:
> I have some DLL errors with JLink on my Raspberry Pi

I saw that earlier when the Jlink was not installed properly, there is also an option to pass the Jlink lib as an argument to pylons (if that is the problem you see)

ldab:
@Kampi 

there is a truck load of unmet dependencies on your raspberry pi 

IMG_0670

ldab:
You need to “sudo apt update” before install

Kampi:
> You need to “sudo apt update” before install

I run it before the installation

image

What I´m wondering about is that running the command `sudo apt-get -y install...` directly on the Pi is resolving and installing all dependencies, but not when it´s executed from the runner.

ldab:
Is raspberry running 64bits OS? The packages that are not installable are “armhf” maybe the hard float is not available on your distro

Kampi:
> Is raspberry running 64bits OS? The packages that are not installable are “armhf” maybe the hard float is not available on your distro

Should be 64 bit because it´s a Raspberry Pi 4. Can check it in the evening when I´m back at home.

ldab:
> > Is raspberry running 64bits OS? The packages that are not installable are “armhf” maybe the hard float is not available on your distro
> 
> Should be 64 bit because it´s a Raspberry Pi 4. Can check it in the evening when I´m back at home.

Maybe try “ sudo dpkg --add-architecture armhf”

jakkra:
Added as part of https://github.com/ZSWatch/ZSWatch/pull/402 and https://github.com/ZSWatch/ZSWatch/pull/402

PR #331, opened 2024-07-30
External (QSPI) flash read optimizations

jakkra:
- 2x FPS when scrolling watchfaces.
- Overall more responsive.
- Few more FPS here and there where QSPI Flash access was the bottleneck.

Thanks to those imporvements there is now no performance issues putting default watchface background images in external flash freeing up 115KB internal flash :partying_face: 

Imporvements:
1. Make all reads from external flash aligned as the QSPI driver will split uniligned reads into 2-3 seperate reads (depending on if both start and end address is unaligned). Now instead make sure all reads are aligned by reading into a temporary buffer.
2. Added caching when reading from external flash. Only one image at a time, but this is fine as LVGL will often read most of one image in a row by row after each other.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -14,6 +14,11 @@
             "command": "west upload_fs --type lfs"
         },
         {
+            "label": "Erase external flash",
+            "type": "shell",
+            "command": "west upload_fs --erase"
+        },
+        {
             "label": "Disable BT",
             "type": "shell",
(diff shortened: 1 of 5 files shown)


Kampi:
Looks good. Feels smooth :)

jakkra:
> Looks good. Feels smooth :)

Good to hear :+1:

PR #335, opened 2024-08-04
Add option for relative battery scale and set limit for battery arc

Kampi:


jakkra:
Can't we do a setting instead? I'd prefer heaviness as much as possible in runtime?

Kampi:
> Can't we do a setting instead? I'd prefer having as much as possible in runtime?
> 
> Can do like this
> 
> https://github.com/jakkra/ZSWatch/blob/fa035e85b102e296bb3cecdd4b6b5e3b30b1f52a/app/src/ui/watchfaces/0_digital/zsw_watchface_digital_ui.c#L464

Yeah of course.

app/prj.conf
```diff
@@ -166,6 +166,7 @@ CONFIG_ZBUS=y
 CONFIG_ZBUS_RUNTIME_OBSERVERS=y
 CONFIG_ZBUS_CHANNEL_NAME=y
 CONFIG_ZBUS_OBSERVER_NAME=y
+CONFIG_ZBUS_CHANNELS_SYS_INIT_PRIORITY=1
```
jakkra, line 169: Why and what is this change for?

app/src/applications/settings/settings_app.c
```diff
@@ -62,6 +63,11 @@ static setting_app_t settings_app = {
     .ble_aoa_enabled = false,
     .ble_aoa_tx_interval = 100,
     .watchface = {
+#if DEBUG
+        .relative_battery = false,
+#else
+        .relative_battery = true,
+#endif
```
jakkra, line 70: Don't think this is much use, once set once, this will be in flash and this value always overwritten. This mat only work once before anything is written to flash I think.

jakkra (approved):
Tested and it works :)

app/prj.conf
Kampi: It´s not a change. I clean up and sort the file a bit. It was on line 206 before.

app/src/applications/settings/settings_app.c
Kampi: Okay I will remove it.

PR #336, opened 2024-08-04
Add extra options for storing Images on the external flash and enable…

ldab:
… it additionally to HW3


jakkra:
Good, just need to double check native_posix is OK.

PR #338, opened 2024-08-06
Add some emojis to notifications

ldab:
Add some emojis to the notification app, it was not possible to add all the emojis as it consumes too much memory. Basically added faces, hand and ⌚

It works for iOS as it seems ANCS send the UTF-8 of the emoji, it seems like gadgetbridge does some pre-processing, need to check if it can be disabled.

- It seems to use another 8KB
- font from https://fonts.google.com/noto/specimen/Noto+Emoji
- added 0x1F601-0x1F64F, 0x1F446-0x1F450,0x231A : can check on http://www.unicode.org/Public/emoji/1.0//emoji-data.txt

I.e:

IMG_0719


Kampi:
Is it possible to move them to the external flash for newer hardware revisions?

jakkra:
> Is it possible to move them to the external flash for newer hardware revisions?

Yes, 

> Is it possible to move them to the external flash for newer hardware revisions?

Yes, will be part of the same change to allow fonts in external flash. But that's another task. Part of it we move this too and can have all emoji.

jakkra:
Can't we keep using same Montserrat font, but set this emoji font ass the fallback font. I think there is a function for the label to set fallback font? So if it's not found in Montserrat, it will search emoji font.

ldab:
> Can't we keep using same Montserrat font, but set this emoji font ass the fallback font. I think there is a function for the label to set fallback font? So if it's not found in Montserrat, it will search emoji font.

this is somewhat set in the font itself: https://github.com/ldab/ZSWatch/blob/ac135bf6e1bc004d03f0f70c774933b2a1c036ba/app/src/images/fonts/noto_emoji_14_light.c#L1598

ldab:
> this is somewhat set in the font itself: https://github.com/ldab/ZSWatch/blob/ac135bf6e1bc004d03f0f70c774933b2a1c036ba/app/src/images/fonts/noto_emoji_14_light.c#L1598

and also here https://docs.lvgl.io/8.2/overview/font.html?highlight=fallback#use-font-fallback

I'll do that

jakkra:
> > Can't we keep using same Montserrat font, but set this emoji font ass the fallback font. I think there is a function for the label to set fallback font? So if it's not found in Montserrat, it will search emoji font.
> 
> this is somewhat set in the font itself: https://github.com/ldab/ZSWatch/blob/ac135bf6e1bc004d03f0f70c774933b2a1c036ba/app/src/images/fonts/noto_emoji_14_light.c#L1598

Ah so the font is actually Montserrat, but with noto emoji. Maybe just change the font name then to reflect that.

ldab:
how bout that? added the emoji as a fallback from Montserrat, it was a const so didn't want to play around to modify it.... it is a bit hidden but it still works.

jakkra:
Very nice, we can merge and then I'll figure out how to get gadgetbridge not convert it and let it stay utf-8

PR #339, opened 2024-08-06
Replace gadget bridge icon when connected to 🍎

ldab:


Kampi:
What is the meaning of this apple :D?

ldab:
IMG_0722

PR #343, opened 2024-08-07
nRF system reset is disabled for rev 5 and above now

Kampi:


jakkra (approved):
Don't really see the need as

Kampi:
> Don't really see the need as

What do you mean?

jakkra:
> > Don't really see the need as
> 
> What do you mean?

Oops sent by mistake.

PR #379, opened 2025-04-02
 hw model v2 + sysbuild + nRF Connect 2.9.1

jakkra:


jakkra:
Not tested on HW v4. but merging anyway. Will fix later if issue.

PR #388, opened 2025-05-15
Fileupload of external FS through mcumgr

jakkra:
Firmware: Created a fake filesystem to be able to upload files through mcumgr.
Website: Support for file upload through mcumgr.

New section at the button for filesystem upload. In future would be nice to just bundle it together with the FW update.
fota


Kampi:
Isn't this done?

jakkra:
> Isn't this done?

Need to fix CI not passing and add some docs, but otherwise yes.

Issue #406, opened 2025-08-11
Issue with Bluetooth Connection on iOS


HongNguyen635:
I'm current on iOS 18.5, and is using nRF Connect Mobile app to connect.
I'm on debug mode, and the log shows that the connection is successful and the Bluetooth icon appeaers on the screen as well, however, time is not updated on the watch. I tried to switch to another phone also on iOS 18 and the same issue persists.

Below is a snipet from the log:
```
[00:00:11.833,312] <inf> ble_comm: Connected <the address> (random)
[00:00:11.833,496] <inf> ble_comm: MTU exchange pending
[00:00:11.833,526] <inf> ble_comm: Interval: 24, latency: 0, timeout: 72
[00:00:11.833,557] <inf> ble_chronos: Bluetooth  Connected
[00:00:11.836,578] <wrn> bt_hci_core: opcode 0x2022 status 0x0c 
[00:00:11.836,608] <err> bt_conn: Failed to set data len (-13)
[00:00:12.094,451] <inf> ble_comm: LE PHY updated: TX PHY LE 2M, RX PHY LE 2M
[00:00:12.154,724] <inf> ble_comm: LE PHY updated: TX PHY LE 2M, RX PHY LE 2M
[00:00:12.184,356] <inf> ble_comm: LE data len updated: TX (len: 251 time: 2120) RX (len: 251 time: 2120)

[00:00:12.184,783] <inf> ble_comm: MTU exchange done. 244
[00:00:12.185,394] <wrn> bt_l2cap: Ignoring data for unknown channel ID 0x003a
[00:00:12.425,292] <dbg> ble_cts: discover_completed_cb: The discovery procedure succeeded```

HongNguyen635:
I'm building with RTC, and with the head up-to-date with main branch.

jakkra:
> I'm building with RTC, and with the head up-to-date with main branch.

You could try without, there is a conf no_rtc.conf to build with, don't think that's the problem, but you could try if you want. Anyway I'll check tonight when I have access to an iPhone.

HongNguyen635:
I'm closing this issue because it's been addressed in #407

PR #407, opened 2025-08-11
ble: fixing ios time not updated

HongNguyen635:
fixing time not updated on ios although bluetooth connection is established. 
Turns out the security level is < `BT_SECURITY_L2`, so my attempt at fixing it is to manually call `bt_conn_set_security`. 
This worked when I test pairing with my iPhone, but unsure if this is the right fix. Still need to test with other iPhone if possible.

Fix for #406

### DIFF
--- a/app/src/ble/ble_comm.c
+++ b/app/src/ble/ble_comm.c
@@ -326,6 +326,33 @@ static void request_mtu_exchange(void)
     }
 }
 
+static void on_pair_request_confirm(bool yes_pressed)
+{
+    if (!yes_pressed) {
+        return;
+    }
+
+    struct bt_conn *conn = NULL;
+    int rc;
+
+    // if the connection is still active, current_con is not null
+    // then we can update the security level
+    if (current_conn) {
+        conn = bt_conn_ref(current_conn);
+    }
+    if (!conn) {
+        LOG_WRN("Pairing accepted but no active connection");
+        return;
+    }
+
+    rc = bt_conn_set_security(conn, BT_SECURITY_L2);
+    if (rc != 0) {
+        LOG_ERR("Failed to set security: %d", rc);
+        bt_conn_disconnect(conn, BT_HCI_ERR_AUTH_FAIL);
+    }
+    bt_conn_unref(conn);
+}
+
 static void ble_connected(struct bt_conn *conn, uint8_t err)
 {
     char addr[BT_ADDR_LE_STR_LEN];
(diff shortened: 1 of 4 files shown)


jakkra:
Yeah the way to do it is to make sure to pair first. Requesting paring may fail here (but not 100 % sure) [as in ble_comm.c ](https://github.com/ZSWatch/ZSWatch/blob/239f6e50d2344673eabc6c7f018b39629b7550b7/app/src/ble/ble_comm.c#L215) as it's setting it to reject pairing. However it could be the case that this code only affects incoming pairing requests, hence your fix solves it. 

Although it may open for an attacker to trigger a pairing with the watch I think.

I'm thinking maybe we keep the warning logs, but don't send the security request. What you think?

HongNguyen635:
Oh that's right I forgot about the security vulnerability, I'll remove the security request.
Making a pop up is a good idea. I can attempt to make the pop up, that'd be much easier to debug if the issue comes from pairing or something else.

HongNguyen635:
I'm wondering when adding the pop up, does the Enable Pairable option in settings still needed. 
I moved this condition in [`ble_comm.c`](https://github.com/ZSWatch/ZSWatch/blob/239f6e50d2344673eabc6c7f018b39629b7550b7/app/src/ble/ble_comm.c#L353) to the button trigger because leaving it there would make it executes before the user presses the event. But then if remove it then enable pairable basically do nothing. I'm unsure which way is the best way.

jakkra:
Great, looks good, let me test it tonight then we can merge it :)

app/src/ble/ble_comm.c
```diff
@@ -349,14 +379,8 @@ static void ble_connected(struct bt_conn *conn, uint8_t err)
     // After some time assume the peer is done and change to longer intervals
     // to save power.
     k_work_schedule(&conn_interval_slow_work, K_MSEC(BLE_COMM_CONN_INT_UPDATE_TIMEOUT_MS));
-
-    if (pairing_enabled) {
```
jakkra, line 353: Don't we need to keep the `if` here, otherwise it will pop up for each time a device connects. I'd say keep the if, but replace the content. We can add this popup I like it, but inside the pairing enabled.

HongNguyen635: sounds good. I updated. Maybe adding a tip pop up to tell user to enable pairing if they want to pair as well?

jakkra (approved):

PR #444, opened 2025-09-18
Sync license texts

Kampi:
Add

```
# Copyright (c) 2025 ZSWatch Project
# SPDX-License-Identifier: Apache-2.0
```

to all `CMakeLists.txt`, `Kconfig` and `.defconfig` files

Add / Replace

```
/*
 * This file is part of ZSWatch project <https://github.com/zswatch/>.
 * Copyright (c) 2025 ZSWatch Project.
 *
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, version 3.
 *
 * This program is distributed in the hope that it will be useful, but
 * WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
 * General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with this program. If not, see <http://www.gnu.org/licenses/>.
 */
 ```
 
 to all `.c` and `.h` files, except those files that have a different license.
 
 @jakkra does this fit for you or shall we change it? Changes should be easy now because we can replace the text on all files which wasn´t possible before.

### CODE REVIEW

File: app/boards/zswatch_legacy_nrf5340_cpuapp_4.conf [hunk 1]
```diff
@@ -1,6 +1,7 @@
-#---------------
```
jakkra, line 1: Replace script not working for this case

app/boards/nrf5340dk_nrf5340_cpuapp.conf
```diff
@@ -1,35 +1,10 @@
-#---------------
```
jakkra, line 1: Replace did not work correctly here

app/pm_static.yml
```diff
@@ -1,6 +1,7 @@
-#-------------------
```
jakkra, line 1: Wrong replacing here too

app/scripts/zswatch_ble_control.py
```diff
@@ -1,6 +1,5 @@
-"""
-Commands over Nordic UART Service
-"""
+# Copyright (c) 2025 ZSWatch Project
```
jakkra, line 1: Same

app/boards/zswatch_legacy_nrf5340_cpuapp_4.conf
Kampi: No this was done manually. I just changed the comment from 
```
#-----------------
# FLASH
#-----------------
```

to 

```
# Flash
```

Shall I undo it?

app/boards/nrf5340dk_nrf5340_cpuapp.conf
Kampi: Same

app/pm_static.yml
Kampi: Same

app/boards/zswatch_legacy_nrf5340_cpuapp_4.conf
jakkra: AH right, no this is fine.

app/boards/nrf5340dk_nrf5340_cpuapp.conf
jakkra: No not the same, in this file a lot of config was removed also. Or was that intentional?

jakkra: Or it's duplicated? And can be removed?

Kampi: It was duplicated. See here

https://github.com/ZSWatch/ZSWatch/blob/main/app/boards/nrf5340dk_nrf5340_cpuapp.conf

PR #578, opened 2026-05-30
Keep timer popup visible until dismissed

david-buxton:
# Summary
- Added a dedicated timer popup with a dismiss button.
- Timer vibration now continues until the popup is dismissed. (note: not tested on hardware)
- Also changed message 'timer triggered' to 'timer is up' as I feel this is a more accurate behaviour description for the user
- Subsequent popup functionality has been tested as follows:

# Testing
Normal (ie non timer) popup still auto-closes correctly via dummy popup (see screenshot)

zsw test image

Also tested:
- Timer popup dismisses correctly
- Triggered timer popup twice in a row.
- Triggered a normal popup after a timer popup.

## Summary by CodeRabbit

* **New Features**
  * Timer alerts now show "Timer X is up!" in a popup with a dedicated Dismiss button.
  * Vibration for timer alerts runs while the popup is shown and stops on dismissal.

* **Bug Fixes**
  * Improved cleanup of timer-related UI so stale auto-close timers and lingering notifications are prevented.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ZSWatch/ZSWatch/pull/578?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/applications/timer/timer_app.c
+++ b/app/src/applications/timer/timer_app.c
@@ -91,9 +91,9 @@ static void alarm_triggered_cb(void *user_data)
     snprintf(buf, sizeof(buf), "Timer %d is up!", timer_id);
 
     /* Run pattern until popup is dismissed. */
-    zsw_vibration_run_pattern_loop(ZSW_VIBRATION_PATTERN_ALARM);
-    zsw_timer_popup_show("Timer", buf, on_close_timer_cb);
-
+    if (zsw_timer_popup_show("Timer", buf, on_close_timer_cb)) {
+       zsw_vibration_run_pattern_loop(ZSW_VIBRATION_PATTERN_ALARM);
+    }
 
 
 }
--- a/app/src/applications/timer/timer_app.c
+++ b/app/src/applications/timer/timer_app.c
@@ -88,11 +88,14 @@ static void alarm_triggered_cb(void *user_data)
     }
 
     static char buf[50];
-    snprintf(buf, sizeof(buf), "Timer %d triggered", timer_id);
+    snprintf(buf, sizeof(buf), "Timer %d is up!", timer_id);
 
-    /* Run pattern until pop up is dismissed or timeout */
+    /* Run pattern until popup is dismissed. */
     zsw_vibration_run_pattern_loop(ZSW_VIBRATION_PATTERN_ALARM);
-    zsw_popup_show("Timer", buf, on_close_timer_cb, 30, false);
+    zsw_timer_popup_show("Timer", buf, on_close_timer_cb);
+
+
+
 }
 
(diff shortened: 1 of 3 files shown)


jakkra:
Looks good, will try on real hw.

jakkra:
One thought, maybe it should only vibrate when it's actually worn? If not then you risk the watch laying somewhere just vibrating forever until battery runs out. What you think? There is some api for this somewhere I think.

david-buxton:
> One thought, maybe it should only vibrate when it's actually worn? If not then you risk the watch laying somewhere just vibrating forever until battery runs out. What you think? There is some api for this somewhere I think.

Hi, good point, though I often use my watch on the table/desk as a timer during exercise/cooking, so it's intentionally not being worn but I would still expect the alert to continue until acknowledged. Worn detection may also lead to false negatives: loose strap, sweaty wrist and/or movement could report not worn even when it is.

It also breaks other valid use cases such charging adjacent use or accessibility reasons. It may also create worse UX than battery drain: missing a timer is often the failure users care about most

Perhaps a better compromise would be to reduce the vibration frequency after a period?

jakkra:
> > One thought, maybe it should only vibrate when it's actually worn? If not then you risk the watch laying somewhere just vibrating forever until battery runs out. What you think? There is some api for this somewhere I think.
> 
> Hi, good point, though I often use my watch on the table/desk as a timer during exercise/cooking, so it's intentionally not being worn but I would still expect the alert to continue until acknowledged. Worn detection may also lead to false negatives: loose strap, sweaty wrist and/or movement could report not worn even when it is.
> 
> It also breaks other valid use cases such charging adjacent use or accessibility reasons. It may also create worse UX than battery drain: missing a timer is often the failure users care about most
> 
> Perhaps a better compromise would be to reduce the vibration frequency after a period?

Yeah that would be better, and maybe stop completely after a few minutes.

david-buxton:
> > > One thought, maybe it should only vibrate when it's actually worn? If not then you risk the watch laying somewhere just vibrating forever until battery runs out. What you think? There is some api for this somewhere I think.
> > 
> > 
> > Hi, good point, though I often use my watch on the table/desk as a timer during exercise/cooking, so it's intentionally not being worn but I would still expect the alert to continue until acknowledged. Worn detection may also lead to false negatives: loose strap, sweaty wrist and/or movement could report not worn even when it is.
> > It also breaks other valid use cases such charging adjacent use or accessibility reasons. It may also create worse UX than battery drain: missing a timer is often the failure users care about most
> > Perhaps a better compromise would be to reduce the vibration frequency after a period?
> 
> Yeah that would be better, and maybe stop completely after a few minutes.

No worries, I'm on it!

PR #583, opened 2026-09-03
Add Altimeter app with altitude tracking and floor counting

elkanamol:
The app calculates altitude from pressure data, applies EMA filtering for stability, detects floors climbed and descended with hysteresis, and publishes altitude through a zbus channel. It also supports periodic calibration, power-management lifecycle handling, and an LVGL interface with altitude history, pressure, calibration status, and floor counters. will close #536 issue. 

## Summary by CodeRabbit

* **New Features**
  * Added an Altimeter app showing current altitude, pressure, calibration status, altitude history, and daily floors climbed or descended.
  * Added automatic location- and weather-based calibration with status updates and retry handling.
  * Added altitude history charting, a runout indicator, and an altimeter icon.
  * Added support for sharing altitude updates with other apps.
* **Documentation**
  * Documented altitude event details, including calibration status and floor counters, in developer documentation and AI guidance.
  * Clarified that altitude readings may not always be calibrated.


app/src/applications/altimeter/altimeter_app.c
```diff
@@ -0,0 +1,393 @@
...
+    ARG_UNUSED(work);
+
+    while (k_msgq_get(&altimeter_work_queue, &work_item, K_NO_WAIT) == 0) {
+        if (work_item.type == ALTIMETER_WORK_CALIBRATION) {
+            alt_state.ref_slp = work_item.value;
+            ui_data.is_calibrated = true;
+            ui_data.has_data = false;
+            ui_data.start_idx = 0;
+            ui_data.valid_count = 0;
+            floor_state.has_last_altitude = false;
```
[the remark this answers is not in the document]
elkanamol: Non issue, this is in purpose. The offered changed, like you will delete user steps because some calibration.

app/src/applications/altimeter/altimeter_app.c
```diff
@@ -0,0 +1,393 @@
...
+
+static void on_zbus_ble_data_callback(const struct zbus_channel *chan)
+{
+    const struct ble_data_event *event = zbus_chan_const_msg(chan);
+
+    if (event->data.type != BLE_COMM_DATA_TYPE_GPS) {
+        return;
+    }
+
+    if (app.current_state != ZSW_APP_STATE_UI_VISIBLE) {
```
jakkra: I'd remove this, and move this check elsewhere where the UI actually gets updated (maybe already checked even). With this check the altitude won't be updated unless the app is open/visibly. The idea with publishing on zbus is to let this app logic run in the background all the time, keeping the altitude up to date and the history up to date.

app/src/images/icons/altimeter_icon.c
```diff
@@ -0,0 +1,101 @@
...
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x76, 0xfc, 0xff, 0xf4, 0x89, 0x1f, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x13, 0x6e, 0xe6, 0xff, 0xff, 0x98, 0x06, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x3b, 0xd7, 0xff, 0xff, 0xfe, 0xca, 0x94, 0x70, 0x5e, 0x68, 0x8c, 0xc1, 0xfb, 0xff, 0xff, 0xe9, 0x59, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x05, 0x61, 0xd2, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xe3, 0x79, 0x0e, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x30, 0x70, 0xa0, 0xbe, 0xd4, 0xc5, 0xa6, 0x7b, 0x3d, 0x03, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+};
+
+const lv_image_dsc_t altimeter_icon32x32 = {
```
jakkra: Rename to just `altimeter_icon` and it will fix the native_sim build

app/src/applications/altimeter/altimeter_app.c
```diff
@@ -0,0 +1,393 @@
...
+static void update_altimeter_ui(struct k_work *work);
+static void process_altimeter_work(struct k_work *work);
+static float get_current_altitude(float pressure_sensor);
+static void check_daily_floor_reset(void);
+static void update_floor_count(float altitude);
+
+// Static variables
+static altimeter_state_t alt_state = { .ref_slp = 1013.25f, .filtered_pressure = 0.0f };
+static floor_count_state_t floor_state = { .last_reset_yday = -1 };
+static float altitude_history[HISTORY_SAMPLES];
```
jakkra: No need for this PR, but more a FYI: you can get persistent history by using https://github.com/ZSWatch/ZSWatch/blob/main/app/src/history/zsw_history.h and it should handle all of it for you.

app/src/applications/altimeter/altimeter_app.c
```diff
@@ -0,0 +1,393 @@
...
+ZBUS_CHAN_DECLARE(pressure_data_chan);
+ZBUS_LISTENER_DEFINE(altimeter_pressure_lis, on_zbus_pressure_callback);
+ZBUS_CHAN_ADD_OBS(pressure_data_chan, altimeter_pressure_lis, 1);
+
+// ZBUS channel this app publishes calibrated altitude to
+ZBUS_CHAN_DECLARE(altitude_data_chan);
+K_WORK_DELAYABLE_DEFINE(altimeter_calibration_work, periodic_calibration_handler);
+K_WORK_DEFINE(ui_update_work, update_altimeter_ui);
+K_WORK_DEFINE(altimeter_work, process_altimeter_work);
+K_MSGQ_DEFINE(altimeter_work_queue, sizeof(altimeter_work_item_t), 8, 4);
```
jakkra: Normally I'm too lazy to use MSGQ for this stuff and just use a static work struct and submit directly to system workqueue as if you get more than one message, you don't really care about that and the latest is the most useful. Also here the update rate is really low, so there will probably never be more than 1 item in the queue ever. But the technique is cleaner than I normally do, but I think you can save some RAM and just make it like 1 or 2? As items will be submitted to this queue only after a HTTP request right? 

EDIT: Just saw it's also submitting pressure data to this queue, how fast are pressure sampled? I think you can ignore my previous comment and this now makes sense.

app/src/applications/altimeter/altimeter_app.c
```diff
@@ -0,0 +1,393 @@
...
+        LOG_ERR("Failed to request GPS for calibration: %d", ret);
+    }
+
+    // Always reschedule so recalibration keeps recurring even if this cycle failed
+    k_work_reschedule(&altimeter_calibration_work, K_SECONDS(ALTIMETER_CALIBRATION_INTERVAL_S));
+}
+
+static void http_rsp_cb(ble_http_status_code_t status, char *response)
+{
+    if (app.current_state != ZSW_APP_STATE_UI_VISIBLE) {
```
jakkra: Same here, remove you just need to not call any UI functions in this function which I don't see any, and it will allow keeping altitude updated in the background every 30min (or whatever interval you put)

jakkra (changes requested):
Really nice! Code looks good, just some behavior I got some comments on. 
Will also test it, but this is code review at least.

app/src/applications/altimeter/altimeter_app.c
elkanamol: @jakkra I don't get your point, we need the msgq or do a lazy static save?
The msgq is because we call the callbacks from the zbus publisher context, so I prefer not to share memory between threads/contexts without protections, and a msgq is simpler then add mutex etc, and also allows me to reduce the workload on the threads and do it in the work_queue context.

jakkra: Yes meant your way is better than my "lazy" way. All good. But wonder if we can reduce the queue size, but depends on how often pressure readings are input.

app/src/applications/altimeter/altimeter_app.c
```diff
@@ -0,0 +1,410 @@
...
+    LOG_INF("GPS received: %.4f, %.4f. Requesting Altimeter data...", lat, lon);
+
+    char http_url[256];
+    snprintf(http_url, sizeof(http_url), HTTP_REQUEST_URL_ALTIMETER_FMT, lat, lon);
+
+    int ret = zsw_ble_http_get(http_url, http_rsp_cb);
+    if (ret != 0 && ret != -EBUSY) {
+        LOG_ERR("Failed to proxy HTTP request: %d", ret);
+        // No async continuation will occur for this cycle, retry sooner
+        reschedule_calibration_work(false);
```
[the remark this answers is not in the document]
elkanamol: @
Sorry, but this exactly example of slope! 

This is part of backof machism added in latest commit, that the calibration must finished even the app is closed, as this app is single point of thrust for altitude and floors data.

jakkra:
You can rebase on main, I fixed a problem that caused Github actions (test workflow) to not check out the code due to external PR.

elkanamol:
Think the issue it from that 'ZSW_LV_IMG_DECLARE(altimeter_icon);' is declared below to app struct. 
I will try to fix it tomorrow.

jakkra:
Just needs a `west astyle ` to fix the formatting issue in ci

elkanamol:
> Just needs a `west astyle ` to fix the formatting issue in ci

Strange!
west format do nothing, I have set auro format on save with astyle per project style. 
I don't try west astyle.

jakkra:
> > Just needs a `west astyle ` to fix the formatting issue in ci
> 
> Strange! west format do nothing, I have set auro format on save with astyle per project style. I don't try west astyle.

Sorry, I wrote the wrong command. `west format` is the correct. Check what astyle version you have, ci runs 3.4.10:
https://github.com/ZSWatch/ZSWatch/blob/2d93427ba620241d84b87ea8d68bc5fccca5f4d6/.github/workflows/build.yml#L53

elkanamol:
> Sorry, I wrote the wrong command. `west format` is the correct. Check what astyle version you have, ci runs 3.4.10:

I checked both the latest 3.6.17 and 3.4.10, both are working well (either manually or automatically with `west format`). The issue is that the document is already formatted (Unchanged), and nothing to do for my files in this PR.
It looks like for me as CI runner issue. 

```
Unchanged  applications\altimeter\altimeter_app.c
Unchanged  applications\altimeter\altimeter_ui.c
Unchanged  applications\altimeter\altimeter_ui.h
```
Any idea? 
Could you check it out and take a closer look?
Meantime, pushing some fake commit (only remove comment).

jakkra:
> > Sorry, I wrote the wrong command. `west format` is the correct. Check what astyle version you have, ci runs 3.4.10:
> 
> I checked both the latest 3.6.17 and 3.4.10, both are working well (either manually or automatically with `west format`). The issue is that the document is already formatted (Unchanged), and nothing to do for my files in this PR. It looks like for me as CI runner issue.
> 
> ```
> Unchanged  applications\altimeter\altimeter_app.c
> Unchanged  applications\altimeter\altimeter_ui.c
> Unchanged  applications\altimeter\altimeter_ui.h
> ```
> 
> Any idea? Could you check it out and take a closer look? Meantime, pushing some fake commit (only remove comment).

Sorry for the confusion, I'll sort it out, you can ignore it for now.

jakkra:
Strange, west format for me formatted altimeter_app:
You can cherry-pick https://github.com/ZSWatch/ZSWatch/commit/177d5d18d31772ed56d9e21d688a8475538a27a2 for now which should fix it

image

jakkra:
Tested it on HW now, it fetches and calibrates, but my height is about -45.4m
`[00:01:34.860,015] <inf> altimeter_app: Calibrated! New MSL Reference: 1018.60 hPa (Temp 2m: 17.9 C)`
My pressure reading is about 1018.66

I'm close to sea level, maybe a bug when pressure > reference?

jakkra:
Ah, probably some smoothing/buffering, now after an hour or so it's 1.4m which makes more sense

elkanamol:
> My pressure reading is about 1018.66

Note that, the MSL from API is not so accurate as the API use gps location with only 3 or 4 digit after the point. So you can have drift, also if you are In a multi-floors building, your pressure sensor should be higher. 
Also, note that the pressure sensor is pretty accurate, the calibration mostly need for environment temperature. 

> Ah, probably some smoothing/buffering, now after an hour or so it's 1.4m which makes more sense

The calibration made in 60:40 ratio for API data. There is smooth transition and gradient with shifting by 0.01 each sample and after while it will drift back to sensor data and recalibrate every 30 min. 
After several try and error this is the most suitable algorithm I found.

Do you think need to modify the algorithm?

EDIT:
> my height is about -45.4m

Did you try to check the results without BLE connection to see if is accurate enough? 
There is defult value for time until get first zbus pressure publish

jakkra:
Sorry for delay, been away over the weekend.

I'm in the 2nd floor of my house, my house is 6m över sea level (according to map from government agency). Walking downstairs does not change much, but maybe the there is a delay in the alg? How "realtime" is it supposed to be?

> Did you try to check the results without BLE connection to see if is accurate enough? There is defult value for time until get first zbus pressure publish

Did not, can let it run today, the issue with running offline is that the sea level pressure can change quite a bit over the day, so it's kindof expected not to work well. 

I think a small change is needed for the initial value, again booting my watch after the weekend it  starts at -60m, If there is no historical data I think it should just pick pure API+latest pressure to set the initial height.

jakkra:
> There is smooth transition and gradient with shifting by 0.01 each sample and after while it will drift back to sensor data and recalibrate every 30 min.

I guess this is there for a reason? Like temporary pressure fluctuation etc.? If so let's keep it, just fix the initial pressure reading starting from default values, instead of the first fetched value and we can merge it. Can improve it more later :)

elkanamol:
> > There is smooth transition and gradient with shifting by 0.01 each sample and after while it will drift back to sensor data and recalibrate every 30 min.
> 
> I guess this is there for a reason? Like temporary pressure fluctuation etc.? If so let's keep it, just fix the initial pressure reading starting from default values, instead of the first fetched value and we can merge it. Can improve it more later :)


As agreed, this design should be enough for this stage, I will open a new issue for next of the design stage including find better smooth algorithm, practice of calibration, enhanced UI and more features that will open by the community.

